PR #18, opened 2025-01-08
Scripts for running

Ale-Coeto:


.gitignore
```diff
@@ -6,4 +6,4 @@
 **/.vscode/
 **/__pychache__/
 .ruff_cache
-
+.env
```
Oscar-gg, line 9: Could you add an .env.example? just in case some .env variables require manual setup

docker/vision/.env
```diff
@@ -0,0 +1,3 @@
+DOCKERFILE=docker/vision/Dockerfile.cpu
```
Oscar-gg, line 1: Hm, doesn't this file match a .gitignore pattern? Was it commited before the .gitignore update?

docker/vision/Dockerfile.cpu
```diff
@@ -0,0 +1,13 @@
+ARG BASE_IMAGE=ubuntu:22.04
+FROM ${BASE_IMAGE}
+
+RUN mkdir -p /workspace/requirements
+WORKDIR /workspace
```
Oscar-gg, line 5: If this dockerfile is based on an image create from Dockerfile.ROS, then I think this line is already executed: https://github.com/RoBorregos/home2/blob/main/docker/Dockerfile.ROS#L54

docker/vision/Dockerfile.cpu
```diff
@@ -0,0 +1,13 @@
...
+
+# Requirements
+COPY vision/vision_general/requirements.txt /workspace/requirements/requirements.txt   
+RUN pip install -r requirements/requirements.txt
+RUN pip install torch==1.11 torchvision==0.12.0
```
Oscar-gg, line 10: is it possible to add these to requirements.txt?

Oscar-gg: nvm, didn't realize that would also affect the Docekrfile.cuda image. Maybe there could be a requirements/cpu.txt and a requirements/cuda.txt, but I think the current implementation is good.

docker/vision/Dockerfile.cuda
```diff
@@ -0,0 +1,31 @@
+ARG BASE_IMAGE=nvidia/cuda:11.4.1-cudnn8-devel-ubuntu20.04
```
Oscar-gg, line 1: good, I think I should also add a default base_image to my docker files 😅. However, if the base image is the specified one, does it include ros 2 installation?

docker/vision/run.sh
Oscar-gg, line 1: Nice

docs/ROS2.md
Oscar-gg, line 1: Nicee!

docker/vision/run.sh
```diff
@@ -0,0 +1,122 @@
...
+#   if [[ "$confirmation" == "y" ]]; then
+#     rm -rf ../../install/ ../../log/ ../../build/
+#     echo "Directories deleted."
+#   else
+#     echo "Operation cancelled."
+#   fi
+# fi
+
+# Setup camera permissions
+if [ -e /dev/video0 ]; then
```
Oscar-gg, line 96: Maybe we could create a separate setup.bash and make a call to it in this script. That way we could reduce the script size and setup the needed permissions when running with other methods (i.e. docker compose commands, or locally)

Oscar-gg (approved):
Lgtm, added some comments.

docker/vision/.env
Ale-Coeto: fixed

docker/vision/Dockerfile.cpu
Ale-Coeto: fixed

docker/vision/Dockerfile.cpu
Ale-Coeto: i think for now one requirements.txt is ok, but if there are more specific req we'll update to diferent files

docker/vision/Dockerfile.cuda
Ale-Coeto: Nop, only ubuntu

PR #22, opened 2025-01-10
Respeaker migration

Oscar-gg:


hri/README.md
```diff
@@ -29,10 +29,15 @@ home2
     └── speech.txt
 ```
 
+
 ## Setup with docker
 
 Run the script `setup.bash` located in `home2/docker/hri` to setup the configuration for docker. The script provides additional setup instructions.
```
GilMM27, line 35: I believe you have to run it each time you restart the host system. Maybe specify that? or correct me if I'm wrong

Oscar-gg: Yes, currently you have to. The expected behavior would be to only run this script once per system, but for some reason one needs to run it after each system restart.

Will add it to the ReadMe.

GilMM27 (approved):
cool, lgtm

PR #23, opened 2025-01-10
Hri nlp

Oscar-gg:
- NLP package setup
- Command interpreter migrated


docker/Dockerfile.ROS
```diff
@@ -50,7 +50,7 @@ RUN curl -sSL https://raw.githubusercontent.com/ros/rosdistro/master/ros.key -o
 RUN apt-get update && apt-get install -y ros-dev-tools
 
 # Setup workspace directory
-RUN mkdir -p /workspace
+RUN mkdir -p /workspace && chown -R $USERNAME:$USERNAME /workspace
```
IvanRomero03, line 53: why is this needed?? could you add some reference here to some docs where we can learn more about this?

docker/hri/setup.bash
```diff
@@ -18,6 +18,11 @@ daemon-binary = /bin/true
...
 
-
 sudo usermod -aG audio $USER # Make sure current user has access to audio resources.
 sudo chmod 777 /dev/snd/* # Allow access to audio devices.
+
+echo "Finished hri setup configuration for docker."
+echo "Note: some of the docker compose files depend on the following environment variables. You may set them on the terminal running the docker compose command, in an .env file in the docker compose directory, or in your ~/.bashrc file:
+
+- export LOCAL_USER_ID=\$(id -u)
+- export LOCAL_GROUP_ID=\$(id -g)"
```
IvanRomero03, line 28: Curious about why not adding the variables into the compose itself or image??? Seems like something very straight forward and automatic, I dont see the need for this to be ran each time by the user if its the same/doesnt have any variability

frida_interfaces/hri/msg/CommandList.msg
```diff
@@ -0,0 +1 @@
+frida_interfaces/Command[] commands
```
IvanRomero03, line 1: Curious, is there a way to add a default value for an attribute in the message? e.g. timestap -> Now() ???

hri/README.md
```diff
@@ -31,15 +31,15 @@ home2
...
 
-- export LOCAL_USER_ID=$(id -u)
-- export LOCAL_GROUP_ID=$(id -g)
+## Using with docker
 
 ```bash
 # Build base image
 # pwd -> home2/docker
-docker compose -f cuda.yaml up
+docker compose -f cuda.yaml build
```
IvanRomero03, line 41: if this is the way to run it and there is no straight forward for the automatic envs above maybe would be worth it to suggest to change this to:
```suggestion
LOCAL_USER_ID=$(id -u)  LOCAL_GROUP_ID=$(id -g) docker compose -f cuda.yaml build
```

hri/packages/nlp/nlp/assets/prompts.py
```diff
@@ -0,0 +1,114 @@
...
+import pytz
+
+CURRENT_CONTEXT = """
+Today is {CURRENT_DATE}.
+Your name is FRIDA (Friendly robotic interactive domestic assistant), a domestic assistant developed by RoBorregos.
+RoBorregos is the representative Robotic team from Tec de Monterrey, Campus Monterrey. It has around 40 members.
+You compete in the Robocup@home competition. Last summer you competed in the Netherlands, at the international competition. Last March you competed in TMR, obtaining 2nd place in Mexico.
+"""
+
+SYSTEM_PROMPT_CI_V2 = """
```
IvanRomero03, line 12: this seems like a nice place to start with the xml prompts, not for now but just so I dont forget lol

hri/packages/nlp/scripts/command_interpreter.py
```diff
@@ -0,0 +1,134 @@
...
+        super().__init__("command_interpreter")
+
+        # Declare and get parameters
+        self.declare_parameter("base_url", "None")
+        self.declare_parameter("model", "gpt-4o-2024-08-06")
+        self.declare_parameter("speech_command_topic", "/speech/raw_command")
+        self.declare_parameter("publish_command_topic", "/task_manager/commands")
+        self.declare_parameter("temperature", 0.5)
+
+        base_url = self.get_parameter("base_url").get_parameter_value().string_value
```
IvanRomero03, line 46: dem, is this the only way for this?? seems like a lot of boilerplate, just curious Im sure this might be the intended way, just seems like a lot (self.declare.. self.get.get.value) lol still seems readable and all, just wondering, it seems like the new thing might be something like pydantic or something like that style

hri/packages/speech/launch/devices_launch.py
```diff
@@ -32,5 +35,13 @@ def generate_launch_description():
...
                 parameters=[speaker_config],
             ),
+            # Node(
+            #     package="speech",
+            #     executable="respeaker.py",
+            #     name="respeaker",
+            #     output="screen",
+            #     emulate_tty=True,
+            #     parameters=[respeaker_config],
+            # ),
```
IvanRomero03, line 45: why not remove?

IvanRomero03: same for above

hri/packages/speech/scripts/respeaker.py
```diff
@@ -0,0 +1,104 @@
...
+            pixel_ring.listen()
+        else:
+            self.get_logger().warn("Command: " + command + " not supported")
+
+
+def main(args=None):
+    rclpy.init(args=args)
+    try:
+        rclpy.spin(Respeaker())
+    except (ExternalShutdownException, KeyboardInterrupt):
```
IvanRomero03, line 97: Add other except for other exceptions maybe?? wouldnt this silence non pacefull shutdowns?

IvanRomero03 (approved):
very nice work!

hri/packages/nlp/scripts/command_interpreter.py
Oscar-gg: I think there are other ways for doing this but this seems to be the intended way, which I think ensures compatibility with ros introspection tools. I do agree that it's a lot of boilerplate 💀. I did a little [reading](https://roboticsbackend.com/rclpy-params-tutorial-get-set-ros2-params-with-python/) and it seems that undeclared parameters could be used, but I guess that it makes it impossible to use stuff like `ros2 param get`.

frida_interfaces/hri/msg/CommandList.msg
Oscar-gg: You can for [simple types](https://docs.ros.org/en/humble/Concepts/Basic/About-Interfaces.html#field-default-value), but I'm unsure if the timestamp -> now() is supported

hri/README.md
Oscar-gg: Hm, I think it would be very inconvenient to type that command each time, unless we use the `run.bash` script or a `Makefile`.

docker/hri/setup.bash
Oscar-gg: hm, good question

Oscar-gg: It can have variability, `id -u` and `id -g` output values based on the current user logged in. It seems that the default UID of the first user created is `1000`, so most people shouldn't have a problem if we hard-code it.
 
I am not sure what would be the best option:

1. We could add an `.env` in the directory of the docker compose files, but that would result in adding many .envs. 
2. We can handle everything on the side of the docker image. That is, at the end of the dockerfiles or somewhere around, we could add `USER ros`. That way we wouldn't need to specify the user manually, however, when mounting the volumes into the docker, one could only edit the files in the dockerfile if the UID specified in the image matches the same as UID of the owner of the files.
3. Add the exports directly to `~.bashrc`. I wouldn't like this, especially if the script could be executed more than once. 

I hadn't tested a lot, but it seems that specifying `user: ${LOCAL_USER_ID}:${LOCAL_GROUP_ID}` doesn't always work. That is, if id -g doesn't match the group specified in the dockerfile, there are several issues:
image

Personally, I think a good option could be 2, and add somewhere that the [UID specified in the image](https://github.com/RoBorregos/home2/blob/main/docker/Dockerfile.ROS#L27) must match the user's `id -u`.

hri/packages/speech/launch/devices_launch.py
IvanRomero03: maybe add uncomment it and add a run flag/parameter??

frida_interfaces/hri/msg/CommandList.msg
IvanRomero03: hard to tell since it seems like the std of types here is small and doesnt have any time form

docker/hri/setup.bash
IvanRomero03: I like the docker image option, adding it to the env or bashrc seems like an invisible option, e.g. a user that doesnt have this config/hasnt read the correct file is going to miss this

hri/packages/speech/scripts/respeaker.py
IvanRomero03: right!

IvanRomero03:
lgtm! 🚀

PR #25, opened 2025-01-13
Manipulacion gpd

GerardoFJ:


IvanRomero03:
What does @Oscar-gg @EmilianoHFlores think about forking the submodules here??? instead of just attaching to an external repo, attach it to a fork of it, not necesarily to be modify by us (even though we could and propose changes) but also to be sure of what version and mantain a stable version for us without the risk of the external dep being removed

EmilianoHFlores:
@IvanRomero03 It is a good idea, I've done that in the old [manipulation](https://github.com/RoBorregos/home-manipulation/tree/main/zed_ws/src) repo to hold the zed-ros dependencies. Just be sure that forks are within the Roborregos org.

IvanRomero03:
Precisly! @EmilianoHFlores being part of the roborregos org is a **must**!

Deivideich:
> What does @Oscar-gg @EmilianoHFlores think about forking the submodules here??? instead of just attaching to an external repo, attach it to a fork of it, not necesarily to be modify by us (even though we could and propose changes) but also to be sure of what version and mantain a stable version for us without the risk of the external dep being removed

Hey, so for you to consider, the main reason to keep a external module as a fork instead of a local version is in case any updates are being done to the Xarm driver and ROS2 connection (RTDE interfaces mainly). It has happened to me with UR arms that after updating their control boxes it is also necessary to keep with the latest changes in their ROS2 driver repos to use their controllers.

Also, when building the Xarm packages it is necessary to download submodules with their respective updated versions to use their SDK, which is the backbone for some of our arm movement related services (https://github.com/xArm-Developer/xarm_ros2/blob/humble/.gitmodules). In case there are any incompatibilities between the SDK and the Xarm ROS2 Driver from the forked version, it might be harder to detect.

Also, as a side note, @GerardoFJ found some launches related to the development of a simulation version compatible with Gazebo ignition that seems to be under development by Ufactory, so eventually it will be necessary to sync the fork with the latest version (so being redundant) or use the external submodule instead

Oscar-gg:
also, pls rename the folder `manipulacion` to `manipulation` to mantain language consistency  🙏

manipulacion/packages/arm_pkg/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
+cmake_minimum_required(VERSION 3.8)
+project(arm_pkg)
+
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
```
IvanRomero03, line 6: can we set this to 
```suggestion
set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

prebuild.sh
```diff
@@ -0,0 +1,89 @@
...
+# Confirm with the user if not auto-accepted
+if [[ "$AUTO_ACCEPT" == false ]]; then
+    read -p "Are you sure you want to proceed? (yes/no): " choice
+    if [[ "$choice" != "yes" ]]; then
+        echo "Aborting script. Please navigate to the correct directory and try again."
+        exit 1
+    fi
+else
+    echo "Auto-accept flag (-y) detected. Proceeding without confirmation."
+fi
```
IvanRomero03, line 42: it would be better if we had an actuall way to check this, this isnt the only place where checking if the running command is in the ws

manipulation/packages/arm_pkg/launch/frida_sim_controller.launch.py
```diff
@@ -0,0 +1,367 @@
...
+        "ros2_control_plugin", default="gazebo_ros2_control/GazeboSystem"
+    )
+
+    add_realsense_d435i = LaunchConfiguration("add_realsense_d435i", default=False)
+    add_d435i_links = LaunchConfiguration("add_d435i_links", default=True)
+    model1300 = LaunchConfiguration("model1300", default=False)
+    robot_sn = LaunchConfiguration("robot_sn", default="")
+    # attach_to = LaunchConfiguration('attach_to', default='base_link')
+    # attach_xyz = LaunchConfiguration('attach_xyz', default='"0.0049756 0 0.34"')
+    # attach_rpy = LaunchConfiguration('attach_rpy', default='"0 0 1.57"')
```
IvanRomero03, line 48: remove?

manipulation/packages/arm_pkg/launch/frida_sim_controller.launch.py
```diff
@@ -0,0 +1,367 @@
...
+
+    # gazebo spawn entity node
+    gazebo_spawn_entity_node = Node(
+        package="gazebo_ros",
+        executable="spawn_entity.py",
+        output="screen",
+        arguments=[
+            "-topic",
+            "robot_description",
+            # '-entity', '{}'.format(xarm_type),
```
IvanRomero03, line 217: unused?

prebuild.sh
GerardoFJ: Ok i'll check it, thanks for the review

manipulation/packages/arm_pkg/src/gpd_test.cpp
IvanRomero03, line 1: for later... can we have some formating for c++ later as the same style implemented for python @Oscar-gg

GerardoFJ: yes

manipulation/packages/arm_pkg/launch/frida_sim_controller.launch.py
GerardoFJ: oh yes sorry, this launch files shouldn't be on the pr

manipulation/packages/arm_pkg/launch/frida_sim_controller.launch.py
GerardoFJ: yes, this are untested launches sorry for the mistake

.github-local/Dockerfile
```diff
@@ -0,0 +1,6 @@
+# ros2-build equivalent as dockerfile for local debugging
+FROM ros:humble AS ros2-build
+
+WORKDIR /ros2_ws
+
+# RUN bash -c "source /opt/ros/humble/setup.sh" && ./.github/workflows/setup.sh && colcon build --symlink-install
```
IvanRomero03, line 6:
```suggestion
```

manipulation/packages/arm_pkg/launch/frida_sim_controller.launch.py
IvanRomero03: dw

prebuild.sh
IvanRomero03: thanks for adding it, its going to be very helpful!

prebuild.sh
```diff
@@ -26,26 +21,24 @@ EOF
 
 # Pause for a few seconds to let the user read the warning
 sleep 5
+SCRIPT_DIR=$(dirname "$(realpath "$0")")
```
IvanRomero03, line 24: nice

IvanRomero03 (approved):
lgtm

IvanRomero03:
Very nice work! Thanks for pulling this up

PR #28, opened 2025-01-15
Node for OpenWakeWord

JocelynVelarde:
Created `kws_oww.py` which is designed for integration with ROS using the OpenWakeWord library for keyword spotting of “No”, “Yes” and “Frida”.


hri/packages/speech/scripts/kws_oww.py
```diff
@@ -0,0 +1,73 @@
...
+import numpy as np
+import openwakeword
+from openwakeword.model import Model
+
+class OpenWakeWordNode(Node):
+    def __init__(self):
+        super().__init__("openwakeword_node")
+        self.get_logger().info("Initializing OpenWakeWord node.")
+
+        self.declare_parameter("model_path", "")
```
IvanRomero03, line 15: is the model going to be included in this path? Maybe we can have a models dir in the package

JocelynVelarde: not very sure where to include the model, if you have any recommendations im very happy to take them

IvanRomero03: its great the one you chose! https://github.com/RoBorregos/home2/blob/893aa0529afcba9c92dbaa79744ca435edb0aada/hri/packages/nlp/launch/nlp_launch.py#L9 checkout the `get_package_share_directory` for the default parameter value

JocelynVelarde: ohhh so should i create a folder named `models/` and then load the models there and then retrieve them from my script using the `get_package_share_directoy`?

IvanRomero03: the `assets/` dir works! and yep using `get_package_share_directoy` will ensure correctness of the path

IvanRomero03 (approved):
other than the ongoing thread, nice work lgtm!

PR #29, opened 2025-01-15
Class for moondream

JocelynVelarde:
Created new utility class `MoondreamWrapper`. The class provides a wrapper around the Moondream model for image processing, including loading images, encoding images, generating captions, querying images, and detecting objects.


Ale-Coeto (approved):
LGTM.
Will need to check the process for downloading the model in a script perhaps

PR #31, opened 2025-01-16
Persist speech socket after reboot

Oscar-gg:


GilMM27 (approved):
works on my machine :)

PR #33, opened 2025-01-17
Subtask manager proposal

Ale-Coeto:
Uses 2 function decorators: 
- Mockable: to return mock data instead of performing the function.
- Check service: to verify if a service is available and return value if not. Will allow task manager to keep going instead of getting stuck for the service being unavailable. For action clients and normal clients.

Takes task_manager as an argument and makes an instance by reference.
- Node methods and attributes will be available through this instance.

When building (in install), Utils is installed in task_manager and subtask_managers to allow testing subtask manager individually (no need to change import path).

Logger standards:
- Services should indicate a request is being made (.info)
- Result should be clearly indicated through logger error or success method (prints in green) 
- No unnecessary logs should be shown (minimal use to show requests and results)
- Warnings or errors should indicate source of error
- This should make debugging easier


task_manager/scripts/subtask_managers/vision_tasks.py
```diff
@@ -54,11 +61,10 @@ def success(self, message) -> None:
         """Print success message"""
         self.node.get_logger().info(f"\033[92mSUCCESS:\033[0m {message}")
 
+    @service_check("save_name_client", -1, TIMEOUT)
```
Oscar-gg, line 64: I think that maybe we need to separate the service checks from the function calls.

I'm not sure, but service checks maybe add more latency to the calls.

Also, maybe a function call could depend on several clients (but that could also indicate that we could separate that function into sub-functions).

Oscar-gg: Also, I think service checking is related to mocking values. I'm unsure how decorators work exactly and their execution order, but my guess is that, in the current implementation, the service is checked and if it passes, then the mocked value is returned (if `self.mock_data` is `true`), right? is we are mocking the value, maybe it is expected that the service isn't available?

task_manager/scripts/utils/decorators.py
```diff
@@ -0,0 +1,62 @@
+"""
```
Oscar-gg, line 1: Very nice  🙏

task_manager/scripts/subtask_managers/vision_tasks.py
```diff
@@ -26,15 +28,20 @@ class VisionTasks:
 
     STATE = {"TERMINAL_ERROR": -1, "EXECUTION_ERROR": 0, "EXECUTION_SUCCESS": 1}
 
-    def __init__(self, task_manager) -> None:
+    def __init__(self, task_manager, mock_data=False) -> None:
         self.node = task_manager
+        self.mock_data = mock_data
```
Oscar-gg, line 33: Maybe it would be good to have more config for the mock data:

- specify which functions we want to mock
- load mock data from other files (for complex types, to avoid adding a lot of mock data to the sub task manager), or to have multiple possible return values

Oscar-gg (approved):

task_manager/scripts/subtask_managers/vision_tasks.py
Ale-Coeto: I was adding the service checks because if one of the services isn't running and a call is made, then the task manager would get stuck. So, if for some reason a service is down, then we can return a value (probably an error status or a mock value) instead of getting stuck. It would also help in cases where the task node is already running and the service node began later, so it would be able to use the service instead of needing to restart the task node. 

Maybe we can add a flag to not check a service if a previous check showed that the service is alive, which would also allow the late service running. This is probably the best approach to avoid delays once the services are running.

For the decorators, they go from closest to the function (bottom to top), so if the mock_data is true, then the services aren't checked, the value is returned instead.

For the dependency on several clients, I hadn't thought of that. If there are cases that can't be separated into sub-functions then we would need a different approach.

task_manager/scripts/subtask_managers/vision_tasks.py
Ale-Coeto: Yeah, I was also thinking about the specifying functions to mock. I think the easiest way would be to add a parameter to the decorator so that if mock_data is True or that parameter is True, then the function is mocked. However, it would be kind of manual and dangerous if we forget one of the functions had the mock on.

Thinking also on the loading data from other files, this could be more complicated and having possible return values as well. I think maybe what could work is to change the decorator to not receive any parameters and use a json to setup the configuration for each function instead. More or less something like this:

```
mock_config = {
    "save_face_name": {"mock": True, "return_value": 100, "delay": 2},
    "another_function": {"mock": False},
}
```
And the decorator wouldn't have parameters `@mockable`
Cons of this would be that it would need to be updated constantly according to each function and is vulnerable to typing errors. I think it is easier to specify the parameters directly on the decorator as it is on top of each function and can be more intuitive, but it would be difficult to have complex return values.

task_manager/scripts/subtask_managers/vision_tasks.py
Oscar-gg: > but I was just checking and it works the other way around xd

Hmm, yes haha. I think it depends on when you call the `func()`, but I would think the parent wrapper gets executed first as it controls the inner functions.

task_manager/scripts/subtask_managers/vision_tasks.py
Oscar-gg: I like the proposed structure for the `mock_config`. I think it would be clean to import that config from somewhere and have that logic separated from the implementations

task_manager/scripts/subtask_managers/vision_tasks.py
Oscar-gg: > task node is already running and the service node began later

Maybe we can make it such that the task node can only start once all the needed services are available.


>Maybe we can add a flag to not check a service if a previous check showed that the service is alive, which would also allow the late service running. This is probably the best approach to avoid delays once the services are running.

I like this idea. If this works, another possibility would be to check all the needed services at the start of the execution. However, both of these approaches assume that once the services are up, they will continue to be for the rest of the execution. I'm unsure if we can assume that :octocat:

Ale-Coeto: Yeah, that's also why i added the check because if a service dies the task manager will get stuck and doesn't log anything so it could be difficult to debug. Anyway we could set the timeout to be very short, but i still think that it is important to make that check to avoid the task manager getting stuck as that would ruin an execution whereas if we return an error then we can handle it and move on.

I think we could leave it like this for now and if it does make things too slow then we could try other approaches

task_manager/scripts/subtask_managers/vision_tasks.py
Ale-Coeto: I would prefer the params because its more direct and intuitive and doesn't need more files. But the config would also be okay. I'll ask to see what the rest think.

PR #34, opened 2025-01-17
added prebuild workflow compability

GerardoFJ:
Workflow changes for manipulation packages build

### DIFF
--- a/.github/workflows/setup.sh
+++ b/.github/workflows/setup.sh
@@ -11,7 +11,10 @@ DEBIAN_FRONTEND=noninteractive
 
 #rosdep install
 DEBIAN_FRONTEND=noninteractive 
-rosdep install --from-paths . -y --ignore-src -r
+rosdep install --from-paths . -y --ignore-src -r --skip-keys gpd
+
+#Running prebuild
+bash ./prebuild.sh -y
 
 # - Importing all dependencies
 colcon build --symlink-install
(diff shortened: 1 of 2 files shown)


.github/workflows/setup.sh
```diff
@@ -11,7 +11,10 @@ DEBIAN_FRONTEND=noninteractive
 
 #rosdep install
 DEBIAN_FRONTEND=noninteractive 
-rosdep install --from-paths . -y --ignore-src -r
+rosdep install --from-paths . -y --ignore-src -r --skip-keys gpd
+
+#Running prebuild
+bash ./prebuild.sh -y
```
IvanRomero03, line 17: I dont see any `prebuild.sh` script

GerardoFJ: is in https://github.com/RoBorregos/home2/pull/25 but i need to add this in order to get the workflow check on the other pr

IvanRomero03: can you move the prebuild.sh script over here??

or maybe this over there

IvanRomero03: and can we have a different workflow that has the gpd so we can actually run the rosdep install and colcon build?

GerardoFJ: If I move this edited workflow to the other PR, will it be checked with the new workflow?

IvanRomero03: yep! it'll run the workflow set on the branch comparing it from, not main

GerardoFJ: ok, so let me remove this pr and branch

PR #37, opened 2025-01-18
Task manager restaurant

Ale-Coeto:


.pre-commit-config.yaml
```diff
@@ -5,8 +5,14 @@ repos:
...
       - id: ruff
+        name: Ruff Global
         types_or: [python, pyi]
+        exclude: ^task_manager/.*
         # args: [--fix]
       # Run the formatter.
+      - id: ruff
+        name: Ruff Taks Manager
+        types_or: [python, pyi]
+        files: ^task_manager/.*
```
Oscar-gg, line 16: Hmm, I think there should be another way of doing this. I recall reading something about it in the docs:

image

https://docs.astral.sh/ruff/configuration/#config-file-discovery

Ale-Coeto: Oh that makes more sense, I'll try it

task_manager/scripts/restaurant_task_manager.py
```diff
@@ -9,45 +9,153 @@
...
-        self.subtask_manager["vision"].find_seat()
-        self.subtask_manager["vision"].detect_person()
-
 
-def main(args=None):
+        while rclpy.ok():
+            if self.current_state == RestaurantTaskManager.TASK_STATES["START"]:
+                Logger.state(self, "Starting task...")
+                self.subtask_manager["hri"].say("I am ready to start the task.")
+                self.current_state += 1
```
Oscar-gg, line 66: I think that this (`self.current_state += 1`) could make it a little less maintainable (that is, if the task states' order happens to change in the future).

task_manager/scripts/subtask_managers/vision_tasks.py
```diff
@@ -27,11 +27,26 @@
...
-
-    def __init__(self, task_manager, mock_data=False) -> None:
+    STATE = {
+        "TERMINAL_ERROR": -1,
+        "EXECUTION_ERROR": 0,
+        "EXECUTION_SUCCESS": 1,
+        "TARGET_NOT_FOUND": 2,
+    }
+    SERVICES = {"detect_person": 0, "save_face_name": 1, "find_seat": 2}
+    SUBTASKS = {
```
Oscar-gg, line 37: I like this! I think it's a very good structure to only check the needed services

task_manager/scripts/subtask_managers/vision_tasks.py
```diff
@@ -44,27 +59,34 @@ def __init__(self, task_manager, mock_data=False) -> None:
...
+                    self.node,
+                    "Find seat service not initialized. (receptionist_commands)",
+                )
+
+        if VisionTasks.SERVICES["detect_person"] in VisionTasks.SUBTASKS[self.task]:
+            if not self.detect_person_action_client.wait_for_server(timeout_sec=TIMEOUT):
+                Logger.warn(
+                    self.node,
+                    "Detect person action server not initialized. (face_recognition)",
+                )
```
Oscar-gg, line 82: hmm, maybe we can find a way of shortening this, as many nodes will probably need to replicate this logic.

task_manager/scripts/subtask_managers/vision_tasks.py
```diff
@@ -27,11 +27,26 @@
 class VisionTasks:
     """Class to manage the vision tasks"""
 
-    STATE = {"TERMINAL_ERROR": -1, "EXECUTION_ERROR": 0, "EXECUTION_SUCCESS": 1}
-
-    def __init__(self, task_manager, mock_data=False) -> None:
+    STATE = {
```
Oscar-gg, line 30: Are these states specifically for vision? Maybe in the future we could move it to a utils file to mantain consistency (when more subtask manager are implemented, oops).

task_manager/scripts/restaurant_task_manager.py
```diff
@@ -9,45 +9,153 @@
...
+        self.subtask_manager["vision"] = VisionTasks(self, task="RESTAURANT", mock_data=MOCK_VISION)
+        self.subtask_manager["nav"] = None
+        self.subtask_manager["manipulation"] = None
+        self.subtask_manager["hri"] = None
+
+        self.detection_attempts = 0
+        self.customers_served = 0
+        self.pan_direction = 1
+        self.coordinates = None
+        self.order = None
```
Oscar-gg, line 52: Nice. I like that all the state is maintained within the task manager. I think this will make it easier to test and debug specific cases.

Oscar-gg (approved):
LGTM 🚀

Added minor regarding the linter.

PR #38, opened 2025-01-20
STT service for hear node

GilMM27:
Implemented a callback groups that allows simultaneous service and subscription callbacks.

Great explanation can be found at: https://discourse.ros.org/t/how-to-use-callback-groups-in-ros2/25255


hri/packages/speech/scripts/useful_audio.py
```diff
@@ -225,9 +222,8 @@ def callback_saying(self, msg):
 
     def callback_keyword(self, msg):
         self.triggered = True
-        if not self.service_active:
-            self.discard_audio()
-            self.compute_audio_state()
+        self.discard_audio()
```
Oscar-gg, line 225: Please confirm that when the service is triggered, a keyword detected wouldn't disrupt the service process.

Oscar-gg (approved):
Lgtm!

hri/packages/speech/scripts/useful_audio.py
GilMM27: fixed it

PR #39, opened 2025-01-20
Refactor/constants

Oscar-gg:
- Added constants package

Use cases:
- Import constants into python files
- Load constants to yaml objects in launch files.

Notes:
- Current constant mechanism only works for python files (we could later create scripts to generate files compatible with cpp files, depending on use cases.)
- Added usage examples in some of HRI config files and launch files.
- Adding constants in `frida_interfaces` was not possible: apparently, `rosidl_generate_interfaces` and `ament_python_install_package` [don't work in the same CMake project](https://docs.ros.org/en/humble/How-To-Guides/Ament-CMake-Python-Documentation.html). There are [several workarounds](https://github.com/ros2/rosidl_python/issues/141) but will leave as it is for the moment.

### DIFF
--- a/hri/packages/speech/launch/devices_launch.py
+++ b/hri/packages/speech/launch/devices_launch.py
@@ -4,17 +4,23 @@ from ament_index_python.packages import get_package_share_directory
 from launch import LaunchDescription
 from launch_ros.actions import Node
 
+from frida_constants import ModuleNames, parse_ros_config
+
 
 def generate_launch_description():
     mic_config = os.path.join(
         get_package_share_directory("speech"), "config", "microphone.yaml"
     )
+
     hear_config = os.path.join(
         get_package_share_directory("speech"), "config", "hear.yaml"
     )
-    speaker_config = os.path.join(
-        get_package_share_directory("speech"), "config", "speaker.yaml"
-    )
+
+    speaker_config = parse_ros_config(
+        os.path.join(get_package_share_directory("speech"), "config", "speaker.yaml"),
+        [ModuleNames.HRI.value],
+    )["say"]["ros__parameters"]
+
     # respeaker_config = os.path.join(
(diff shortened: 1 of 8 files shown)


IvanRomero03 (approved):
fk it we ball

hri/packages/speech/launch/devices_launch.py
```diff
@@ -4,17 +4,23 @@
...
 def generate_launch_description():
     mic_config = os.path.join(
         get_package_share_directory("speech"), "config", "microphone.yaml"
     )
+
     hear_config = os.path.join(
         get_package_share_directory("speech"), "config", "hear.yaml"
     )
-    speaker_config = os.path.join(
-        get_package_share_directory("speech"), "config", "speaker.yaml"
```
IvanRomero03, line 16: im not getting why are we changing this??

Oscar-gg: yaml config has keys with values "REPLACE". `parse_ros_config` updates such values, using the constants defined in `ModuleNames.HRI.value`

IvanRomero03: lmao, seems like a lot isnt it?

IvanRomero03: I would prefer to have all in one place, why should we keep the input name in the config?

IvanRomero03: but seems ok either way, this also seems pretty elegant, I dont dislike it

IvanRomero03: feel free to close this, just wanted to know why

Oscar-gg: > I would prefer to have all in one place, why should we keep the input name in the config?

Hmm, so that we know which values to insert / which values are being replaced?

PR #40, opened 2025-01-22
Subtaskmanager config from files

Oscar-gg:
- Load config objects from files
- Validate config objects with `pydantic`
- Implemented metaclass for `subtask_managers` to mock functions
- Added `task_manager/scripts/test_manager.py` and `task_manager/scripts/config/hri/debug.py` for usage example

### DIFF
--- a/task_manager/package.xml
+++ b/task_manager/package.xml
@@ -14,6 +14,8 @@
   <depend>std_msgs</depend>
   <depend>geometry_msgs</depend>
   <depend>frida_interfaces</depend>
+  <depend>frida_constants</depend>
+
 
   <test_depend>ament_lint_auto</test_depend>
   <test_depend>ament_lint_common</test_depend>
@@ -22,4 +24,3 @@
     <build_type>ament_cmake</build_type>
   </export>
 </package>
-
--- a/task_manager/scripts/config/hri/debug.py
+++ b/task_manager/scripts/config/hri/debug.py
@@ -24,7 +24,7 @@ __config = {
     ],
     "mock_config": [
         {
-            "function_name": "extract_dat",
+            "function_name": "extract_data",
             "enabled": True,
             "mock_data": mock_extract_data,
         },
--- a/task_manager/scripts/config/hri/debug.py
+++ b/task_manager/scripts/config/hri/debug.py
@@ -1,0 +1,45 @@
+from random import randint
+
+from utils.config import SubtaskConfig
+
+from frida_constants.hri_constants import COMMAND_INTERPRETER_SERVICE, HEAR_SERVICE
+
+
+def mock_extract_data(query, complete_text):
+    """Pick a random word from complete_text"""
+
+    split_text = complete_text.split()
+
+    return split_text[randint(0, len(split_text) - 1)]
+
+
+__config = {
+    "topic_config": [
+        {
+            "topic_name": COMMAND_INTERPRETER_SERVICE,
+            "enabled": False,
+            "type": "service",
+        },
+        {"topic_name": HEAR_SERVICE, "enabled": False, "type": "service"},
+    ],
+    "mock_config": [
+        {
+            "function_name": "extract_dat",
+            "enabled": True,
+            "mock_data": mock_extract_data,
+        },
+        {
+            "function_name": "say",
+            "enabled": True,
+            "mock_data": "Succeeded!",
+        },
+        {
+            "function_name": "hear",
+            "enabled": True,
+            "mock_data": "Hi Frida, can you bring me a glass of water?",
+        },
+    ],
+    "strict": True,
+}
+
+config = SubtaskConfig(**__config)
--- a/task_manager/scripts/subtask_managers/hri_tasks.py
+++ b/task_manager/scripts/subtask_managers/hri_tasks.py
@@ -1,0 +1,79 @@
+#!/usr/bin/env python3
+
+"""
+HRI Subtask manager
+"""
+
+from typing import Union
+
+import rclpy
+from rclpy.node import Node
+from subtask_meta import SubtaskMeta
+
+from frida_constants.hri_constants import (
+    COMMAND_INTERPRETER_SERVICE,
+    DATA_EXTRACTOR_SERVICE,
+    HEAR_SERVICE,
+    SPEAK_SERVICE,
+)
+from frida_interfaces.srv import STT, CommandInterpreter, ExtractInfo, Speak
+
+TIMEOUT = 5.0
+
+
+class HRITasks(metaclass=SubtaskMeta):
+    """Class to manage the vision tasks"""
+
+    STATE = {"TERMINAL_ERROR": -1, "EXECUTION_ERROR": 0, "EXECUTION_SUCCESS": 1}
+
+    def __init__(self, task_manager, config) -> None:
+        self.node = task_manager
+
+        self.speak_service = self.node.create_client(Speak, SPEAK_SERVICE)
+        self.hear_service = self.node.create_client(STT, HEAR_SERVICE)
+        self.extract_data_service = self.node.create_client(
+            ExtractInfo, DATA_EXTRACTOR_SERVICE
+        )
+
+        self.command_interpreter_service = self.node.create_client(
+            CommandInterpreter, COMMAND_INTERPRETER_SERVICE
+        )
+
+    def say(self, text: str, now: bool = False) -> None:
+        """Method to publish directly text to the speech node"""
+        self.node.get_logger().info(f"Sending to saying service: {text}")
+
+        self.speak_service(text)
+
+    def extract_date(self, query, complete_text) -> str:
+        pass
+
+    def hear(self, timeout: float) -> str:
+        pass
+
+    def interpret_keyword(self, keyword: str, timeout: float) -> str:
+        pass
+
+    def refactor_sentence(self, sentence: str) -> str:
+        pass
+
+    def find_closest(self, query: str, options: Union[list[str], str]) -> str:
+        pass
+
+    def ask(self, question: str) -> str:
+        """Method to publish directly text to the speech node"""
+        pass
+
+    def command_interpreter(self, text: str) -> str:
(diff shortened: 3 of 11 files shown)


task_manager/package.xml
```diff
@@ -22,4 +24,3 @@
     <build_type>ament_cmake</build_type>
   </export>
 </package>
-
```
GerardoFJ, line 25: Xd

task_manager/package.xml
```diff
@@ -14,6 +14,8 @@
   <depend>std_msgs</depend>
   <depend>geometry_msgs</depend>
   <depend>frida_interfaces</depend>
+  <depend>frida_constants</depend>
+
```
GerardoFJ, line 18: Xd

task_manager/scripts/config/hri/debug.py
```diff
@@ -0,0 +1,45 @@
...
+        {
+            "function_name": "hear",
+            "enabled": True,
+            "mock_data": "Hi Frida, can you bring me a glass of water?",
+        },
+    ],
+    "strict": True,
+}
+
+config = SubtaskConfig(**__config)
```
IvanRomero03, line 45:
```suggestion
if name==main
config = SubtaskConfig(**__config)
```

task_manager/scripts/subtask_managers/subtask_meta.py
```diff
@@ -0,0 +1,64 @@
...
+                        # Replace original function with mock
+                        mock_count += 1
+                        if callable(mock_data):
+                            mock_method = Mock(side_effect=mock_data)
+                            setattr(instance, method_name, mock_method)
+                        else:
+
+                            def create_mock_method(mock_data, method_name_):
+                                def mock_method(*method_args, **method_kwargs):
+                                    Logger.mock(instance.node, method_name_)
```
IvanRomero03, line 38: can we instace the logger, not calling this class?

task_manager/scripts/utils/config.py
IvanRomero03, line 1: are  these being used??

task_manager/package.xml
Oscar-gg: xd

task_manager/scripts/config/hri/debug.py
Oscar-gg: I think we need to leave it like this, otherwise we wouldn't be able to import the config from other files.

task_manager/scripts/utils/config.py
Oscar-gg: yes, in `debug.py`, to validate the configuration's shape

task_manager/scripts/subtask_managers/subtask_meta.py
Oscar-gg: current `Logger` only contains `@staticmethod`s, we would need to modify that first

task_manager/scripts/utils/config.py
IvanRomero03: I cannot see where in debug.py is being used

Oscar-gg: `from utils.config import SubtaskConfig`

IvanRomero03: lmao nice

IvanRomero03 (approved):
lgtm thanks for explaining

task_manager/scripts/subtask_managers/hri_tasks.py
```diff
@@ -0,0 +1,79 @@
...
+
+        self.speak_service = self.node.create_client(Speak, SPEAK_SERVICE)
+        self.hear_service = self.node.create_client(STT, HEAR_SERVICE)
+        self.extract_data_service = self.node.create_client(
+            ExtractInfo, DATA_EXTRACTOR_SERVICE
+        )
+
+        self.command_interpreter_service = self.node.create_client(
+            CommandInterpreter, COMMAND_INTERPRETER_SERVICE
+        )
```
GerardoFJ, line 40: This topic name declaration could be a great example of using ROS 2 parameters. I'm thinking of a testing situation where I might want to change the service name. In that case, I would have to modify the entire constants file. This is precisely the purpose of using parameters in a node.

task_manager/scripts/subtask_managers/hri_tasks.py
```diff
@@ -0,0 +1,79 @@
...
+    def find_closest(self, query: str, options: Union[list[str], str]) -> str:
+        pass
+
+    def ask(self, question: str) -> str:
+        """Method to publish directly text to the speech node"""
+        pass
+
+    def command_interpreter(self, text: str) -> str:
+        pass
+
```
GerardoFJ, line 69: Unfinished functions?

task_manager/scripts/test_manager.py
GerardoFJ, line 1: This can be related with something like this? [https://docs.ros.org/en/humble/Tutorials/Intermediate/Testing/Testing-Main.html]url

GerardoFJ (commented):
Check comments

task_manager/scripts/subtask_managers/hri_tasks.py
Oscar-gg: yes, its the subtaskmanager skeleton (if someone else wants to mock it with the added functionalitites)

task_manager/scripts/test_manager.py
Oscar-gg: Not necessarily. This test manager was to showcase the use of the new functionalities. The link you provided has to do with automatic tests.

task_manager/scripts/subtask_managers/hri_tasks.py
Oscar-gg: > In that case, I would have to modify the entire constants file

Well, you would need to modify only the value of that 1 constant (which would update references across the codebase)

task_manager/scripts/subtask_managers/hri_tasks.py
```diff
@@ -0,0 +1,79 @@
...
+HRI Subtask manager
+"""
+
+from typing import Union
+
+import rclpy
+from rclpy.node import Node
+from subtask_meta import SubtaskMeta
+
+from frida_constants.hri_constants import (
```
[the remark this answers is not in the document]
Oscar-gg: Interesting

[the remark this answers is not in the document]
Oscar-gg: Is there a way of getting parameters without the need of having a node?

for instance, do something like:

`SPEAK_SERVICE = rclpy.get_param("param_name")`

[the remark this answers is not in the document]
Oscar-gg: Either way, if we wanted to use ros2 params as constants, there should be a node that always sets them, right?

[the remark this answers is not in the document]
GerardoFJ: > Is there a way of getting parameters without the need of having a node?
> 
> for instance, do something like:
> 
> `SPEAK_SERVICE = rclpy.get_param("param_name")`

why would you want to use a parameter  outside a node?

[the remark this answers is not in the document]
GerardoFJ: > Either way, if we wanted to use ros2 params as constants, there should be a node that always sets them, right?

If you want to use parameters as constants, refer to the YAML configuration file system. Check out this article: [https://roboticsbackend.com/ros2-yaml-params/](url)

task_manager/scripts/subtask_managers/hri_tasks.py
GerardoFJ: But this affects all the nodes using these constants. What if I'm trying to isolate some tests and avoid using or referring to specific topics?

task_manager/scripts/subtask_managers/hri_tasks.py
GerardoFJ: ok nice work

task_manager/scripts/test_manager.py
GerardoFJ: nice, i see the work here

task_manager/scripts/subtask_managers/hri_tasks.py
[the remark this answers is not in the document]
Oscar-gg: > > Is there a way of getting parameters without the need of having a node?
> > for instance, do something like:
> > `SPEAK_SERVICE = rclpy.get_param("param_name")`
> 
> why would you want to use a parameter outside a node?

True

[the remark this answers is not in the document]
Oscar-gg: > > Either way, if we wanted to use ros2 params as constants, there should be a node that always sets them, right?
> 
> If you want to use parameters as constants, refer to the YAML configuration file system. Check out this article: [https://roboticsbackend.com/ros2-yaml-params/](url)

Agree, I've used this and works great. However, how can we make sure that the constants specified in several `yaml` files are the same? For instance, let's say that `Node A` launches a service with topic `B`. How could we make sure that in `Node C`, the client is instantiated with the same topic `B`?

task_manager/scripts/subtask_managers/hri_tasks.py
Oscar-gg: > But this affects all the nodes using these constants.

Yes, the point is to maintain consistency.

> What if I'm trying to isolate some tests and avoid using or referring to specific topics?

You could change the topic manually (i.e, change the variable name)

IvanRomero03: > But this affects all the nodes using these constants. What if I'm trying to isolate some tests and avoid using or referring to specific topics?

then you would be risking the other packages

IvanRomero03: the concept and a core thing to keep in mind is that in order to pull off a hard thing as this proyect, we need to ensure consistency and redundancy within all the codebase, ros params only serve their purposes in the "Runtime" space

PR #45, opened 2025-02-02
Feature/embeddings

lleo-13:
chromadb locally + node item categorizations with 4 services: add_item, remove_item, update_item,query_item. Edit: chromadb as microservice and build_embeddings as a service


hri/packages/embeddings/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
+cmake_minimum_required(VERSION 3.8)
```
Oscar-gg, line 1: maybe we could add this to the nlp package (to group related functionality)? But I guess this should also be fine

hri/packages/embeddings/chromadb/0c3a54f7-2348-4ed0-99e6-79dc300b131d/data_level0.bin
Oscar-gg, line 1: I think we should probably not commit these files?

Oscar-gg: (Add to .gitignore)?

hri/packages/embeddings/scripts/item_categorization.py
```diff
@@ -0,0 +1,186 @@
...
+
+            response.success = True
+            response.message = "Items added successfully"
+        except Exception as e:
+            response.success = False
+            response.message = f"Failed to add items: {str(e)}"
+            self.get_logger().error(response.message)
+        return response
+
+    def remove_item_callback(self, request, response):
```
Oscar-gg, line 72: Nice! I think it's very good to have CRUD functionality implemented

hri/packages/embeddings/scripts/item_categorization.py
```diff
@@ -0,0 +1,186 @@
...
+        self.query_item_service = self.create_service(QueryItem, 'query_item', self.query_item_callback)
+
+        # Initialize ChromaDB client
+        script_dir = Path(__file__).resolve().parent
+        chroma_path = str(script_dir / "../chromadb")
+        self.chroma_client = chromadb.PersistentClient(path=chroma_path)
+
+        # Configure the embedding function
+        self.sentence_transformer_ef = embedding_functions.SentenceTransformerEmbeddingFunction(
+            model_name="all-MiniLM-L12-v2"
```
Oscar-gg: maybe we can add the model as a parameter, and leave this as the default value. See [say.py](https://github.com/RoBorregos/home2/blob/main/hri/packages/speech/scripts/say.py) for an example of ros parameter.

hri/packages/embeddings/scripts/item_categorization.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def _get_or_create_collection(self, collection_name):
+        """Helper method to get or create a collection"""
+        try:
+            return self.chroma_client.get_or_create_collection(name=collection_name)
+        except Exception as e:
+            self.get_logger().error(f"Error retrieving collection '{collection_name}': {str(e)}")
+            raise
+
+    def build_embeddings(self):
```
Oscar-gg, line 168: Is this executed only once or each time the node is launched? If this takes significant time to execute, maybe we can check if the db already contains the information and skip if applicable.

Oscar-gg (approved):
LGTM, great work!

PR #46, opened 2025-02-04
Removed branch tag on GPD package and specified branch on Xarm package in .gitmodules

Deivideich:
GPD package doesnt need a specified branch, it causes a crash when doing the cloning process

Instead, the Xarm ROS2 pkg needs its branch to be specified at humble, otherwise the colcon build process will fail


Oscar-gg:
LGTM 🚀

Oscar-gg (approved):
Lgtm 🚀

PR #49, opened 2025-02-10
Openwakeword corrections

JocelynVelarde:
### Fixes
* Modified the model loading mechanism to dynamically load all models from a specified directory instead of a single model file.
* Implemented a cooldown mechanism to prevent rapid repeated wakeword detection by using `last_detection_time` and `detection_cooldown`.

### Dependency updates:

* Added `openwakeword==0.6.0` to the `requirements/speech.txt` file to ensure the required version of the OpenWakeWord library is installed.

> Let me know if there are more modifications needed, would love the feedback!

### DIFF
--- a/hri/packages/speech/scripts/kws_oww.py
+++ b/hri/packages/speech/scripts/kws_oww.py
@@ -78,7 +78,8 @@ class OpenWakeWordNode(Node):
                     self.get_logger().info(
                         f"Wakeword '{keyword}' detected with score {scores[-1]:.2f}"
                     )
-                    self.publisher.publish(Bool(data=True))
+                    detection_info = {"keyword": keyword, "score": scores[-1]}
+                    self.publisher.publish(detection_info)
                     self.last_detection_time = current_time
                 break
 
--- a/hri/packages/speech/scripts/kws_oww.py
+++ b/hri/packages/speech/scripts/kws_oww.py
@@ -4,6 +4,8 @@ from rclpy.node import Node
 from std_msgs.msg import Bool
 from frida_interfaces.msg import AudioData
 from openwakeword.model import Model
+import os
+import time
 
 
 class OpenWakeWordNode(Node):
@@ -18,6 +20,7 @@ class OpenWakeWordNode(Node):
         self.declare_parameter("audio_topic", "/rawAudioChunk")
         self.declare_parameter("detection_topic", "/wakeword_detected")
         self.declare_parameter("chunk_size", 1280)
+        self.declare_parameter("detection_cooldown", 1.0)
 
         model_path = self.get_parameter("model_path").get_parameter_value().string_value
         inference_framework = (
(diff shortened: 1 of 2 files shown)


hri/packages/speech/scripts/kws_oww.py
```diff
@@ -32,14 +35,22 @@ def __init__(self):
...
-        # Initialize OpenWakeWord
+        # Handle model loading dynamically from a directory
         if model_path:
-            self.get_logger().info(f"Loading model from {model_path}")
-            self.oww_model = Model(
-                wakeword_models=[model_path],
-                inference_framework=inference_framework,
-            )
+            self.get_logger().info(f"Loading models from {model_path}")
+            model_files = [f for f in os.listdir(model_path) if f.endswith(".onnx")]
```
Oscar-gg, line 45: Great!

hri/packages/speech/scripts/kws_oww.py
```diff
@@ -50,25 +61,25 @@ def __init__(self):
         self.publisher = self.create_publisher(Bool, detection_topic, 10)
         self.create_subscription(AudioData, audio_topic, self.audio_callback, 10)
 
+        # Flag to prevent rapid repeated publishing
+        self.last_detection_time = time.time() - self.detection_cooldown
```
Oscar-gg, line 65: I'm not entirely sure if it would be a good idea but maybe we could have a last detection time per keyword

hri/packages/speech/scripts/kws_oww.py
```diff
@@ -50,25 +61,25 @@ def __init__(self):
...
         for keyword, buffer in self.oww_model.prediction_buffer.items():
             scores = list(buffer)
-            if (
-                scores[-1] > 0.5
-            ):  # NOTE: This value can be decreased for more sensitive yet less effective detection
-                self.get_logger().info(
-                    f"Wakeword '{keyword}' detected with score {scores[-1]:.2f}"
-                )
-                self.publisher.publish(Bool(data=True))
+            if scores[-1] > 0.5:
```
Oscar-gg, line 74: It would be good if we could configure the threshold per keyword (however, I'm also not sure if it's needed as I haven't experimented much and seen if the threshold changes a lot per keyword)

hri/packages/speech/scripts/kws_oww.py
```diff
@@ -50,25 +61,25 @@ def __init__(self):
...
+                # Only publish if the score is above a certain threshold
+                current_time = time.time()
+                if current_time - self.last_detection_time >= self.detection_cooldown:
+                    self.get_logger().info(
+                        f"Wakeword '{keyword}' detected with score {scores[-1]:.2f}"
+                    )
+                    self.publisher.publish(Bool(data=True))
```
Oscar-gg, line 81: Could you please publish the keyword detected? [Here](https://github.com/RoBorregos/home2/blob/main/hri/packages/speech/scripts/useful_audio.py#L237) is an example of a string publisher.

Oscar-gg (changes requested):

hri/packages/speech/scripts/kws_oww.py
JocelynVelarde: so an individual timing for each word?

hri/packages/speech/scripts/kws_oww.py
JocelynVelarde: to make a good detection of the word, the developers reccommended 0.5, but if you want it to be more accurate yet less sensitive perhaps 0.7 is the smarter choice, and all the values basically vary in 0.1 ~ 0.2 when someone speaks

hri/packages/speech/scripts/kws_oww.py
JocelynVelarde: thanks will do

JocelynVelarde: done!

hri/packages/speech/scripts/kws_oww.py
Oscar-gg: okok, sounds good. do you know if there are variations for larger and smaller words? For example, using smaller thresholds for larger words and bigger thresholds for smaller words? E.g. if the word is "RoBorregos" maybe we could lower the threshold, but if the word is "yes" maybe we could elevate the threshold?

Oscar-gg (approved):
lgtm

PR #61, opened 2025-02-14
Manipulation and images for CUDA development

EmilianoHFlores:
Includes:

- Sample pipeline to deploy containers with and without CUDA
- Expansion of `./run.sh` method introduced for `vision` containers
- Changed container structure to allow keeping `install` `log` and `build` folders within docker memory

Requires:

- Testing on only CPU PCs

No Jetson image is provided yet.


GerardoFJ (approved):
LGTM :rocket: !!!!!!!

PR #64, opened 2025-02-18
GPD_INSTALL_LIB variable is updated in .bashrc everytime the prebuild.sh is executed

Deivideich:
Now in case it is necessary to redo the prebuild in other directory it wont be stopped due to older versions in .bashrc


Oscar-gg (approved):
lgtm, asdfjasdfasfadsf

PR #76, opened 2025-02-21
Moondream class in python

JocelynVelarde:
Created multiple methods for moondream class

- Missing to create ROS node, because we need to define the methods that we wan’t to use.


Ale-Coeto (approved):
LGTM

PR #81, opened 2025-02-22
Pymoveit2 interface for xArm6

EmilianoHFlores:
Contains the frida_pymoveit2 package with examples to call moveit functions from external scripts


JLDominguezM (approved):
good

PR #85, opened 2025-02-23
Merge commits with person cropper

JocelynVelarde:
Created method using YOLO to crop the largest area person. 
Passed that cropped image to moondream description generator.


vision/vision_general/scripts/receptionist_commands.py
```diff
@@ -207,6 +208,56 @@ def detect_person(self):
...
+                (int(x + w / 2), int(y + h / 2)),
+                (0, 255, 0),
+                2,
+            )
+            cropped_frame = frame[
+                int(y - h / 2) : int(y + h / 2), int(x - w / 2) : int(x + w / 2)
+            ]
+            return cropped_frame
+
+        if (time.time() - self.start_time) > CHECK_TIMEOUT:
```
Ale-Coeto, line 255: I think this is not necessary because the detect_and_crop_person is not running with a timer, its just a normal function that runs once. 
The timers are usually used to run loops in ros2 because a regular while, for example, will stop the node from listening to publishers, so the `create_timer(time, function)` is used instead to begin running the same function every x miliseconds, which can be stopped using the timer.cancel()

Ale-Coeto (commented):
Looks very good, just check pls the timer comment

vision/vision_general/scripts/receptionist_commands.py
JocelynVelarde: alright, so then i should delete all the timing things?

Ale-Coeto: yes pls

JocelynVelarde: done!

Ale-Coeto (approved):
LGTM

IvanRomero03 (commented):
very nice! srry im late to the review but great work!

PR #87, opened 2025-02-25
Add remove plane service, initial iter

IvanRomero03:
Todo:
- [ ] Filter points by height and maybe other xyz constrains
- [ ] Add objects logic


afr2903 (commented):
lgtm 🚀

afr2903 (approved):
image

Deivideich (changes requested):
In remove_plane.cpp file please change the node name since it says test_node.

Deivideich:
Also, when possible, please continue improving the functionality of filterPointsByHeight function with support for multiple ranges of heights to ignore. 

The same way, in removePlane function please add the distance threshold and max iterations as paramters, with the actual values as default

PR #110, opened 2025-03-05
Identification between: standing, laying down, sitting

JocelynVelarde:
- Added moondream method to detect person pose between the 3 different categories
- Modified the query parameters for the beverage location


vision/vision_general/scripts/moondream_lib.py
```diff
@@ -32,57 +32,25 @@ def generate_person_description(self, encoded_image, query, stream=False):
...
-        horizontal_position = "left" if bbox_center_x < image_center_x else "right"
-        vertical_position = "up" if bbox_center_y < image_center_y else "down"
 
-        return horizontal_position, vertical_position
-
-
-# Example usage
+        if not detect_result["objects"]:
+            return "not found"
+        else:
```
JocelynVelarde, line 38: I need some revision checking if the logging is correct in here for the `receptionist_commands.py`

vision/vision_general/scripts/receptionist_commands.py
```diff
@@ -155,21 +165,41 @@ def person_description_callback(self, request, response):
...
+        if response.location == "not found":
+            self.get_logger().warn("Beverage not found")
+            response.success = False
+            return response
+        else:
+            response.location = self.moondream_model.find_beverage(
+                encoded_image, request.beverage
+            )
+            self.success(f"Beverage location: {response.location}")
+            return response
```
JocelynVelarde, line 185: Over here is the usage of the `find_beverage()`

vision/vision_general/requirements.txt
```diff
@@ -1,9 +1,9 @@
 opencv-python
 pathlib
 tqdm
-numpy<2
+numpy>=2.1.2
```
IvanRomero03, line 4: this seems problematic, are we sure this is ok for the other deps in the package?? I think we usually use 1.22 or 1.24 but if this is the new way to go, just to know

vision/vision_general/scripts/moondream_lib.py
IvanRomero03: @Ale-Coeto

Ale-Coeto: I think the enum proposal would be a good approach

vision/vision_general/scripts/receptionist_commands.py
```diff
@@ -155,21 +165,41 @@ def person_description_callback(self, request, response):
...
         self.get_logger().info("Executing service Beverage Location")
-        beverage = request.beverage
 
         if self.image is None:
             response.location = "No image received yet."
             return response
 
         frame = self.image
         encoded_image = self.moondream_model.encode_image(frame)
-        result = self.moondream_model.find_beverage(encoded_image, beverage)
```
Ale-Coeto, line 175: I think the model should be called before using the response

```suggestion

  result = self.moondream_model.find_beverage(
      encoded_image, request.beverage
  )
    
  if result == "not found":
    response.success = False
    response.location = "not found"
    self.get_logger().warn("Beverage not found")
  
  else:
    response.success = True
    response.location = result
    self.success(f"Beverage location: {response.location}")
    
return response
                
```

vision/vision_general/scripts/receptionist_commands.py
```diff
@@ -155,21 +165,41 @@ def person_description_callback(self, request, response):
...
+        query = "Determine if the person is sitting, standing, or lying down. Just mention the pose, no aditional information is needed."
+        cropped_frame = self.detect_and_crop_person()
+        encoded_image = self.moondream_model.encode_image(cropped_frame)
+
+        # Use same method for dperson_description but with different query
+        response.description = self.moondream_model.generate_person_description(
+            encoded_image, query, stream=False
         )
-        self.success(f"Beverage location: {response.location}")
         return response
```
Ale-Coeto, line 203: For feedback on the response you could add the response.success value (this would also require a change on the BeverageLocation.srv)

frida_interfaces/vision/srv/BeverageLocation.srv
```diff
@@ -1,3 +1,4 @@
+bool request
 string beverage
 ---
 string location
```
Ale-Coeto, line 4: i would recommend adding a bool for success

vision/vision_general/scripts/receptionist_commands.py
Ale-Coeto, line 185: I think the model should be called before using the response

```suggestion

  result = self.moondream_model.find_beverage(
      encoded_image, request.beverage
  )
    
  if result == "not found":
    response.success = False
    response.location = "not found"
    self.get_logger().warn("Beverage not found")
  
  else:
    response.success = True
    response.location = result
    self.success(f"Beverage location: {response.location}")
    
return response
                
```

vision/vision_general/scripts/receptionist_commands.py
Ale-Coeto, line 203: For feedback on the response you could add the response.success value (this would also require a change on the BeverageLocation.srv)

vision/vision_general/requirements.txt
JocelynVelarde: yes! that is the version required to run moondream, it can be 2.1.2 or >3

IvanRomero03: Nice

JocelynVelarde:
@Ale-Coeto fixed all comments 
- about enum declaration
- correct `find_beverage() `method calling in `receptionist_commands`
- Success messages using the `.srv `declaration

pls review 🤓

vision/vision_general/scripts/receptionist_commands.py
```diff
@@ -34,15 +35,16 @@
 CAMERA_TOPIC = "/zed/zed_node/rgb/image_rect_color"
 CHECK_PERSON_TOPIC = "/vision/detect_person"
 PERSON_DESCRIPTION_TOPIC = "/vision/person_description"
+PERSON_POSTURE_TOPIC = "/vision/person_posture"
 BEVERAGE_TOPIC = "/vision/beverage_location"
 FIND_SEAT_TOPIC = "/vision/find_seat"
 IMAGE_TOPIC = "/vision/img_person_detecion"
```
IvanRomero03, line 41: not for the scope of this pr, but these should be in frida_constans @Ale-Coeto ...

IvanRomero03 (approved):
nice

PR #111, opened 2025-03-05
Added motion planning action server

EmilianoHFlores:
- Moved `Planner` objects to a new package `frida_motion_planning`
- Made pymoveit2 use exclusive to the `motion_planning_server` node.
- Include services to send pose and joint goals, get joint positions
- Modified manipulation subtask manager to handle new action server


afr2903 (approved):
lgtm 🫂

PR #118, opened 2025-03-06
Fixed demo follow_face

GerardoFJ:
Fixed manipulation subtask
Fixed task manager
changed names


task_manager/launch/demo_face.launch.py
IvanRomero03, line 1: can we rename this as demo_follow_face.launch.py

IvanRomero03 (commented):
are we sure these are all the needed things in the node??

IvanRomero03 (approved):
lgtm

PR #123, opened 2025-03-08
LLM unit testing

Oscar-gg:


hri/packages/nlp/config/extract_data.yaml
```diff
@@ -1,6 +1,6 @@
 extract_data:
   ros__parameters:
-    # base_url: "http://localhost:11434/v1"
-    # model: "llama3.2"
+    base_url: "http://localhost:11434/v1"
+    model: "deepseek-r1:7b"
```
afr2903, line 4: apoko si pa el deepseek?
pls verify it's actually better for the application, don't add it just for the hype

hri/packages/nlp/nlp/assets/dialogs.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def get_common_interests_dialog(
+    person1Name, person2Name, person1Interests, person2Interests
+):
+    return {
+        "messages": [
+            {
+                "role": "system",
+                "content": "You will be presented with the interests of two people, your task is to get the common interests between them. Give a short answer with one common interest.",
```
afr2903, line 11: adding an example within the system prompt may increase the success rate: and example of the interests of two people and the correct answer of one in common, especially useful if an specific format is expected

hri/packages/nlp/test/test_data_extractor.py
```diff
@@ -0,0 +1,68 @@
...
+    ("Maria's favorite fruit is mango", "fruit", ExtractedData(data="mango").model_dump_json()),
+    ("The Eiffel Tower is located in Paris", "city", ExtractedData(data="Paris").model_dump_json()),
+    ("We watched a great movie called Inception", "movie", ExtractedData(data="Inception").model_dump_json()),
+    ("Carlos loves programming in Python", "language", ExtractedData(data="Python").model_dump_json()),
+    ("The capital of France is Paris", "capital", ExtractedData(data="Paris").model_dump_json()),
+    ("We traveled to Japan last summer", "country", ExtractedData(data="Japan").model_dump_json()),
+    ("His favorite sport is basketball", "sport", ExtractedData(data="basketball").model_dump_json()),
+    ("Sarah lives in New York City", "location", ExtractedData(data="New York City").model_dump_json()),
+    ("My best friend’s name is Daniel", "friend", ExtractedData(data="Daniel").model_dump_json()),
+]
```
afr2903, line 48: I'd add cases where it should return empty.

hri/packages/nlp/nlp/assets/schemas.py
```diff
@@ -0,0 +1,7 @@
+from typing import Optional
+
+from pydantic import BaseModel
+
+
+class ExtractedData(BaseModel):
+    data: Optional[str] = None
```
afr2903, line 7: If the response for a case where it's not possible to extract data it's an empty string `""`,  remove the Optional, it will be easier to catch the Pydantic exception from here than later in the process

afr2903 (approved):
check the suggestions, lgtm 🚀

hri/packages/nlp/nlp/assets/dialogs.py
```diff
@@ -0,0 +1,92 @@
+from .schemas import ExtractedData
```
afr2903, line 1: try to avoid relative imports 🙏

afr2903 (approved):
relative imports 😵

hri/packages/nlp/config/extract_data.yaml
Oscar-gg: Yes. It was the only local llm that I tested and that passed all test cases...

hri/packages/nlp/nlp/assets/dialogs.py
Oscar-gg: :o
Interesting, do you suggest an alternative?

hri/packages/nlp/nlp/assets/dialogs.py
Oscar-gg: got it! since there are currently no test cases for the `common_interest`, I will make sure to modify the system prompt once the cases are implemented (to increase similarity to expected outputs as much as possible).

hri/packages/nlp/test/test_data_extractor.py
Oscar-gg: got it. added one atm (`("", "drink", ExtractedData(data="").model_dump_json()`)

hri/packages/nlp/nlp/assets/schemas.py
Oscar-gg: done!

hri/packages/nlp/nlp/assets/dialogs.py
Oscar-gg: will wait for your suggestion to possibly make changes before merging 
image

afr2903: I usually run into conflicts trying to solve relative imports xd
But the import should be like:
```python
from nlp.nlp.assets.schemas import ExtractedData
```
If you simply change the import to this, it may prompt an error of type `ModuleNotFoundError`.

The first requirement for a python module it's having an empty `__init__.py` in the root folder of the module (`nlp`).

Then, the solution that worked for me when I kept getting the `ModuleNotFoundError` was:

Create a file name `conftest.py` at the root module folder (`nlp`) with the following content (props to Gemini Thinking):
```python
# tests/conftest.py or conftest.py at project root
import sys
import os

# Add the project root to the Python path
sys.path.insert(0, os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
```

Oscar-gg: Modified relative path to `from nlp.assets.schemas import ExtractedData` and seems to be working in `test_data_extractor.py` and when running with ros. Will leave like that atm.

PR #154, opened 2025-03-13
Octomap working in real arm

Deivideich:
Now it is possible to launch the real arm with the octomap using the Moveit's built in Octomap Plugin for OA

The main change was the addition of the .yaml with the plugin description, and the octomap yaml loader inside the move_group node, as seen in frida_moveit_common.launch

use: ros2 launch arm_pkg frida_moveit_config.launch.py 

and ros2 launch zed_wrapper zed_camera.launch.py camera_model:=zed2 publish_tf:=false

configure the octomap resultion in the move_group and node and its own .yaml inside config folder


IvanRomero03 (approved):

EmilianoHFlores (approved):
ok

PR #156, opened 2025-03-13
xArm mode set

EmilianoHFlores:
Includes adjusting xArm state and mode to ensure compatibility when sending goals through MoveIt.

Bonus, includes dualshock demo with HRI speaking


GerardoFJ (approved):
LGTM

PR #157, opened 2025-03-14
Addition of a simple gazebo environment

Deivideich:


Deivideich:
Addition of a simple gazebo world with a shelf and a table with an object

EmilianoHFlores (approved):
very good

PR #160, opened 2025-03-15
Sets of joints and move single joint example

EmilianoHFlores:
Methods for receptionist, added named sets of joints and an example to call it alongside moving a single joint


GerardoFJ (approved):
LGTM :rocket:

PR #165, opened 2025-03-15
Receptionist mar15

Ale-Coeto:


task_manager/scripts/follow_face_node.py
```diff
@@ -4,21 +4,27 @@
 Task Manager for Demo follow face
 """
 
+from time import time as t
```
Oscar-gg, line 7: 🧐

task_manager/scripts/follow_face_node.py
```diff
@@ -52,70 +63,161 @@ def __init__(self):
...
 
+    def set_state(self):
+        Logger.info(self, "Activating arm")
+
+        # Set state
+        state_request = SetInt16.Request()
+        state_request.data = 0
+        # Set mode
+        # 0: position control mode
+        mode_request = SetInt16.Request()
```
Deivideich, line 101: As seen in the line 16 the service that is going to be called is the xarm joint velocity service which needs mode 4 and state 0, as seen in https://github.com/xArm-Developer/xarm_ros2/tree/humble/xarm_api in robot mode 4, so please change the mode_equeste value to 4..

task_manager/scripts/subtask_managers/vision_tasks.py
```diff
@@ -206,14 +206,14 @@ def find_drink(self, drink: str, timeout: float = TIMEOUT):
 
             if not result.success:
                 Logger.warn(self.node, "No drink found")
-                return self.STATE["TARGET_NOT_FOUND"], ""
+                return "not found", self.STATE["TARGET_NOT_FOUND"]
 
         except Exception as e:
             Logger.error(self.node, f"Error finding drink: {e}")
-            return self.STATE["EXECUTION_ERROR"]
+            return "not found", self.STATE["EXECUTION_ERROR"]
```
Oscar-gg, line 213: Interesting, I like this way of returning status

Deivideich (changes requested):
The service /xarm/vc_set_joint_velocity needs to be in state 0 and mode 4, and in the set_state function the requested mode is 0 (position mode). In case the velocity service is called, it will not work.

task_manager/scripts/receptionist_task_manager.py
```diff
@@ -208,16 +218,30 @@ def run(self):
...
+                joint_positions = self.subtask_manager.manipulation.get_joint_positions(
+                    degrees=True
+                )
+                joint_positions["joint1"] = joint_positions["joint1"] - seat_angle
+                self.subtask_manager.manipulation.move_joint_positions(
+                    joint_positions=joint_positions, velocity=0.5, degrees=True
+                )
                 # self.subtask_manager.manipulation.pan_to(seat_angle)
-                angle = self.subtask_manager.vision.find_seat()
+                status, angle = self.subtask_manager.vision.find_seat()
```
Oscar-gg, line 233: I think we should always use the same order. I think this pr modifies the order to res, status in some of the functions

task_manager/scripts/subtask_managers/vision_tasks.py
Oscar-gg: Is the status also returned in the @mockable? Would the expected behavior be to return a `MOCKET_STATE` or to be able to mock the status?

task_manager/scripts/subtask_managers/manipulation_tasks.py
```diff
@@ -9,6 +9,7 @@
 import rclpy
 from rclpy.node import Node
 from utils.logger import Logger
+
 from xarm_msgs.srv import SetInt16, SetInt16ById, MoveVelocity
```
Oscar-gg, line 13: Not related to this pr specifically, but is it possible to migrate these messages to `frida_interfaces`? That way we could avoid building the `xarm_msgs` package for every container that wants to run the task managers. Is the xarm_msgs package fast to build or does it depend on `rosdep install`?  @EmilianoHFlores @JLDominguezM

Oscar-gg (commented):
lgtm

task_manager/scripts/follow_face_node.py
Ale-Coeto: @IvanRomero03

IvanRomero03: needs testing from our end

IvanRomero03: please dont hold this pr over this

IvanRomero03 (approved):
lgtm

task_manager/scripts/follow_face_node.py
Deivideich: @IvanRomero03 PRs should only be approved if they have been properly tested. Approving this without the requested change would make the debugging process difficult, as no explicit feedback from the Xarm is provided regarding this error.

Deivideich (approved):

task_manager/scripts/receptionist_task_manager.py
Oscar-gg: ?

PR #174, opened 2025-03-17
Temporary workaround to avoid the robot moving by itself when is stationary in the gazebo simulation

Deivideich:
Temporary change of wheel's weight from 0.5 kgs to 100 kgs so the robot wont move when stationary. Further research is required to have a more realistic performance


IvanRomero03 (approved):
lgtm @GerardoFJ

PR #185, opened 2025-03-20
Added dialog for known places with schema

JocelynVelarde:
Added test file, schema and dialog for room identification from known places list: living room, dinning room, kitchen, bedroom, entrance.


Oscar-gg (commented):
lgtm, please check comments

PR #187, opened 2025-03-20
Vision poses

Yair-Reyes:
Add pose detection and gesture detection

### DIFF
--- a/vision/vision_general/vision_general/pose_detection.py
+++ b/vision/vision_general/vision_general/pose_detection.py
@@ -17,12 +17,12 @@ class PoseDetection:
         results = self.pose.process(image_rgb)
         if results.pose_landmarks:
             landmarks = results.pose_landmarks.landmark
-            left_hip = landmarks[mp_pose.PoseLandmark.LEFT_HIP]
-            right_hip = landmarks[mp_pose.PoseLandmark.RIGHT_HIP]
-            left_knee = landmarks[mp_pose.PoseLandmark.LEFT_KNEE]
-            right_knee = landmarks[mp_pose.PoseLandmark.RIGHT_KNEE]
-            left_ankle = landmarks[mp_pose.PoseLandmark.LEFT_ANKLE]
-            right_ankle = landmarks[mp_pose.PoseLandmark.RIGHT_ANKLE]
+            left_hip = landmarks[self.mp_pose.PoseLandmark.LEFT_HIP]
+            right_hip = landmarks[self.mp_pose.PoseLandmark.RIGHT_HIP]
+            left_knee = landmarks[self.mp_pose.PoseLandmark.LEFT_KNEE]
+            right_knee = landmarks[self.mp_pose.PoseLandmark.RIGHT_KNEE]
+            left_ankle = landmarks[self.mp_pose.PoseLandmark.LEFT_ANKLE]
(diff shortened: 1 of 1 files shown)


vision/vision_general/vision_general/pose_detection.py
```diff
@@ -13,10 +13,48 @@ def __init__(self):
...
+        if results.pose_landmarks:
+            landmarks = results.pose_landmarks.landmark
+            left_hip = landmarks[self.mp_pose.PoseLandmark.LEFT_HIP]
+            right_hip = landmarks[self.mp_pose.PoseLandmark.RIGHT_HIP]
+            left_knee = landmarks[self.mp_pose.PoseLandmark.LEFT_KNEE]
+            right_knee = landmarks[self.mp_pose.PoseLandmark.RIGHT_KNEE]
+            left_ankle = landmarks[self.mp_pose.PoseLandmark.LEFT_ANKLE]
+            right_ankle = landmarks[self.mp_pose.PoseLandmark.RIGHT_ANKLE]
+
+            if left_hip.y < left_knee.y and right_hip.y < right_knee.y:
```
EmilianoHFlores, line 27: Add "standing" and others as constants so we know how to interpret them in other scripts/nodes

vision/vision_general/vision_general/pose_detection.py
```diff
@@ -13,10 +13,48 @@ def __init__(self):
...
+            left_wrist = landmarks[self.mp_pose.PoseLandmark.LEFT_WRIST]
+            right_wrist = landmarks[self.mp_pose.PoseLandmark.RIGHT_WRIST]
+            left_elbow = landmarks[self.mp_pose.PoseLandmark.LEFT_ELBOW]
+            right_elbow = landmarks[self.mp_pose.PoseLandmark.RIGHT_ELBOW]
+            left_shoulder = landmarks[self.mp_pose.PoseLandmark.LEFT_SHOULDER]
+            right_shoulder = landmarks[self.mp_pose.PoseLandmark.RIGHT_SHOULDER]
 
-    def detectGesture(self):
-        pass
+            if left_wrist.y < left_shoulder.y and right_wrist.y < right_shoulder.y:
```
EmilianoHFlores, line 47: same

vision/vision_general/vision_general/pose_detection.py
```diff
@@ -13,10 +13,48 @@ def __init__(self):
         self.mp_drawing = mp.solutions.drawing_utils
 
     def detectPose(self, image):
-        pass
+        image_rgb = cv2.cvtColor(image, cv2.COLOR_BGR2RGB)
+        results = self.pose.process(image_rgb)
```
EmilianoHFlores, line 17: we run results both in detectPose and in detectGesture, are they run at the same time? or are they in different nodes? If we run both then we can save one self.pose.process.

EmilianoHFlores (approved):
ok

vision/vision_general/vision_general/pose_detection.py
Ale-Coeto: They are in the same node, but it is unlikely they will run at the same time. It is usually either get pose or get gesture

vision/vision_general/vision_general/pose_detection.py
Ale-Coeto: Could an enum be a better option or not really necessary?

Ale-Coeto (approved):
LGTM

PR #192, opened 2025-03-22
fixes

GilMM27:
* stopped keywords from deleting previous audio
* service now returns empty when the audio state returns to idle
* increased vad threshold to transition to idle faster


Oscar-gg (approved):
lgtm!!

PR #200, opened 2025-03-24
Object detection handler

Chapa-1810:
Object detection handler:
Main purpose is to avoid processing of detections within nodes. Capable of returning a list of all objects with a determined label or the closest object if depth and 3d point stamped is enabled


vision/object_detection_handler/LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
IvanRomero03, line 190: seems uncomplete papu

vision/object_detection_handler/src/object_detection_handler.cpp
EmilianoHFlores, line 1: great

vision/object_detection_handler/src/object_detection_handler.cpp
```diff
@@ -0,0 +1,220 @@
...
+    void declare_parameters(){
+      this->declare_parameter("DETECTIONS_TOPIC", "/detections");
+      this->declare_parameter("TARGET_FRAME", "BASE_LINK");
+      this->declare_parameter("TRANSFROM_POINTS", false);
+      this->declare_parameter("VERBOSE", false);
+      this->declare_parameter("IOU_THRESHOLD", 0.75);
+      this->declare_parameter("RECORDED_SECONDS", 5);
+    }
+
+    void get_parameters(){
```
IvanRomero03, line 76: nice its def the most elegant way I've seen so far

vision/object_detector_2d/config/parameters.yaml
EmilianoHFlores, line 1: lots of topics here, please add them too to frida_constants and maybe we can find a way to make them match without editing >1 file??

vision/object_detector_2d/launch/object_detector_node.launch.py
EmilianoHFlores, line 1: would it be better to have the handler here too? the fewer terminals the better

vision/object_detector_2d/scripts/object_detector_node.py
```diff
@@ -0,0 +1,431 @@
...
+
+        # Frames per second throughput estimator
+        self.fps = None
+        callFpsThread = threading.Thread(target=self.callFps, args=(), daemon=True)
+        callFpsThread.start()
+
+        if self.node_params.VERBOSE:
+            self.get_logger().info("Object Detector 2D Node has been started")
+
+    def set_parameters(self):
```
EmilianoHFlores, line 107: D:

vision/object_detector_2d/scripts/object_detector_node.py
EmilianoHFlores, line 1: This code seems rather long for what it does, can we have the visualization on a different area? I believe the methods use to visualize 3d marker arrays may also be used in other packages/areas so they can make use of them too. Not for this PR

EmilianoHFlores (changes requested):
ok! Check the topics and add them to frida_constants so we can use them easily on external code

EmilianoHFlores (changes requested):
ok! Check the topics and add them to frida_constants so we can use them easily on external code

vision/object_detection_handler/LICENSE
Chapa-1810: Added owner (me), should I added as RoBorregos?

vision/object_detection_handler/config/parameters.yaml
```diff
@@ -0,0 +1,8 @@
+object_detection_handler:
+  ros__parameters:
+    TRANSFROM_POINTS : True
+    VERBOSE : False
+    DETECTIONS_TOPIC : /detections
```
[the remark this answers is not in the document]
Chapa-1810: Fixed this

vision/object_detector_2d/config/parameters.yaml
Chapa-1810: Added them to frida constants, both for python and cpp

vision/object_detector_2d/scripts/object_detector_node.py
Chapa-1810: :+1:

vision/object_detector_2d/scripts/object_detector_node.py
Chapa-1810: Probably some of this functions/methods could be added to vision_utils file.

EmilianoHFlores (approved):
ok

PR #201, opened 2025-03-25
Vision moondream server

Ale-Coeto:
Added moondream server using grpc
Added docker compose for both node container and server container
Update moondream_node to use grpc service


docker/vision/docker-compose.yaml
```diff
@@ -29,4 +29,7 @@ services:
     tty: true
     stdin_open: true    
     working_dir: /workspace
-    command: ["/bin/bash", "-i", "-l"]
\ No newline at end of file
+    command: ["/bin/bash", "-i", "-l"]
+
+# include:
```
Oscar-gg, line 34: Maybe remove?

Oscar-gg (approved):
LGTM

PR #209, opened 2025-03-26
Update subtask manager for vision and added tracking service

Ale-Coeto:
Added Status class for subtask managers to return a status response.
Added person tracking to subtask manager.

### DIFF
--- a/task_manager/scripts/utils/status.py
+++ b/task_manager/scripts/utils/status.py
@@ -1,7 +1,8 @@
 from enum import Enum
 
+
 class Status(Enum):
-    TERMINAL_ERROR=-1
-    EXECUTION_ERROR=0
-    EXECUTION_SUCCESS=1
-    TARGET_NOT_FOUND=2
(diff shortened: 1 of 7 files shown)


Oscar-gg (approved):
LGTM

task_manager/scripts/utils/status.py
```diff
@@ -0,0 +1,8 @@
+from enum import Enum
+
+
+class Status(Enum):
+    TERMINAL_ERROR = -1
+    EXECUTION_ERROR = 0
+    EXECUTION_SUCCESS = 1
+    TARGET_NOT_FOUND = 2
```
Oscar-gg, line 8: Should we add status SERVICE_CHECK, MOCKED, and TIMEOUT?

Ale-Coeto: I was also thinking of the service check, but it would probably have the same implications of an execution or terminal error. If we have many status it could become hard to handle each case

PR #210, opened 2025-03-26
Added HRI run.sh

Oscar-gg:


docker/hri/run.sh
```diff
@@ -0,0 +1,143 @@
...
+
+
+# Check if the container exists
+EXISTING_CONTAINER=$(docker ps -q -f name=$SERVICE_NAME)
+if [ -z "$EXISTING_CONTAINER" ]; then
+    echo "No container with the name $SERVICE_NAME exists. Building and starting the container now..."
+    if [ $ENV_TYPE == "cpu" ]; then
+        docker compose -f docker-compose-cpu.yml up --build -d
+    elif [ $ENV_TYPE == "gpu" ]; then
+        docker compose -f docker-compose-cpu.yml up --build -d
```
Ale-Coeto, line 109: Should it be gpu or is cpu correct?

docker/hri/run.sh
```diff
@@ -0,0 +1,143 @@
...
+if [ -z "$TASK" ]; then
+    docker exec $SERVICE_NAME /bin/bash
+# else
+#     if [ -z "$detached" ]; then
+#         docker compose exec $SERVICE_NAME bash -c "$SETUP && $RUN"
+#     else
+#         echo "Running in detached mode..."
+#         docker compose exec -d $SERVICE_NAME bash -c "$SETUP && $RUN"
+#     fi
+fi
```
Ale-Coeto, line 143: If no task flag is sent then this will open a bash terminal, is this correct?

docker/hri/run.sh
Oscar-gg: cpu is correct. Currently, it uses the same for both cpu and gpu in laptop

docker/hri/run.sh
Oscar-gg: yes

PR #212, opened 2025-03-26
Chroma/minor changes

lleo-13:


lleo-13:
-Retrieve locations with coordinates as metadata(optional).
-Retrieve items with locations as metadata(optional).
-Refactor of query method. Add context on each entry for better embedding performance.

frida_interfaces/hri/srv/AddItem.srv
```diff
@@ -1,7 +1,6 @@
 # AddItem.srv
 string document 
-string collection
-string metadata
```
Oscar-gg, line 4: why was the metadata removed?

frida_interfaces/hri/srv/QueryItem.srv
```diff
@@ -1,10 +1,9 @@
 # QueryItem.srv
 string query  # The search query
-string collection
```
Oscar-gg, line 3: why was the collection removed? Does it perform the search in all collections?

frida_interfaces/hri/srv/QueryItem.srv
```diff
@@ -1,10 +1,9 @@
...
 int16 topk #number of results 
-string where
+bool return_location #return location?
```
Oscar-gg, line 4: If we have the QueryLocation.srv, maybe we can remove this form here and use the other service for the location calls.

Oscar-gg (commented):
Good work! Please check the comments. If possible, it would be great to add tests to verify it works for a variety of cases

Oscar-gg (approved):
LGTM! Make sure to resolve merge conflicts before merging

PR #215, opened 2025-03-27
Status Refactor + ask and confirm implementation

Oscar-gg:
- Refactored to use return states from enum
- Added routine to confirm a question and to ask a question and confirm its answer
- Added `test_hri_manager.py`


task_manager/scripts/test_hri_manager.py
Ale-Coeto, line 1: i think we could have a test manager and add it to the gitignore

task_manager/scripts/subtask_managers/hri_tasks.py
```diff
@@ -39,22 +39,20 @@
...
 
 
-class ReturnStates(Enum):
-    TERMINAL_ERROR = -1
-    EXECUTION_ERROR = 0
-    EXECUTION_SUCCESS = 1
-    SERVICE_CHECK = 2
-    TIMEOUT = 3
+def confirm_query(interpreted_text, target_info):
+    return f"Did you said {target_info}?"
```
Ale-Coeto, line 51: Small change to Did you say {}.

task_manager/scripts/subtask_managers/hri_tasks.py
```diff
@@ -39,22 +39,20 @@
...
 
 
-STATE = {"TERMINAL_ERROR": -1, "EXECUTION_ERROR": 0, "EXECUTION_SUCCESS": 1, "SERVICE_CHECK": 2}
+def confirm_interpretation(interpreted_text, target_info):
+    return f"Did you said {interpreted_text}?"
```
Ale-Coeto, line 55: Same

Ale-Coeto (approved):
LGTM

task_manager/scripts/subtask_managers/hri_tasks.py
Oscar-gg: Oh right, mb

task_manager/scripts/test_hri_manager.py
Oscar-gg: Hm, personally I would like to have it in the repo to test in my laptop and the orin (without having to copy paste to much).

Is there a reason why we should gitignore it?

Ale-Coeto: Just to avoid having several test managers

Oscar-gg: Refactored the test manager to use flags

PR #217, opened 2025-03-31
feat: use moondream with ollama

Oscar-gg:
Test using ollama and moondream

Steps to set up:
- Use hri's `ollama.yaml` docker compose
- Run `ollama run moondream`
- You can now make requests to the endpoint referenced in the script


IvanRomero03:
ollama runs on gpu, right?

Oscar-gg:
> ollama runs on gpu, right?

Not necessarily. If it doesn't find a gpu I think it defaults to cpu. In this case it did run on gpu

PR #232, opened 2025-04-04
Vision tracker

Ale-Coeto:
Update tracker to publish 3d normalized point and debugged for running in orin.


vision/packages/vision_general/scripts/tracker_node.py
```diff
@@ -61,6 +92,7 @@ def setup(self):
             "backward": None,
             "left": None,
             "right": None,
+            "coordinates": [],
```
IvanRomero03, line 95: any issue on converting this dict to a pydantic model?

IvanRomero03 (approved):
lgftm

vision/packages/vision_general/scripts/tracker_node.py
Ale-Coeto: I'll check it for a future iteration

PR #233, opened 2025-04-04
Hotfix: Constants + Collision Objects

EmilianoHFlores:
Fixed 2 important issues:

- Some constants were missing on main branch
- Collision objects added to object for pick were overlapping with plane collision, making it impossible to move the robot after attaching them


IvanRomero03 (approved):
lgtm

PR #243, opened 2025-04-05
Zed and gripper working on simulation

EmilianoHFlores:
Includes

- Working zed and gripper from real robot on simulation
- Working object detector 3d pixel deprojection


Deivideich (approved):
It works :+1:

PR #247, opened 2025-04-09
Moondream lib test

Ale-Coeto:
Made moondream services more general for queries:
Query general image
Query on a person (detects first and then crops the image)
Query by crop (send bbox to crop and then query - example: for bag description)


JocelynVelarde (approved):
👍

PR #248, opened 2025-04-09
Integration of help me carry task manager initial order into main branch

Deivideich:


task_manager/scripts/help_me_carry.py
```diff
@@ -0,0 +1,188 @@
...
+        "CONFIRM_FOLLOWING": 2,
+        "FOLLOWING_TO_DESTINATION": 3,
+        "ASK_TO_POINT_THE_BAG": 4,
+        "DETECT_THE_BAG": 5,
+        "RECEIVE_THE_BAG": 6,
+        "GRASP_THE_BAG": 7,
+        "RETURN_TO_STARTING_LOCATION": 8,
+        "PLACE_THE_BAG": 9,
+        "END": 10,
+    }
```
IvanRomero03, line 33: can we have this as an enum?

task_manager/scripts/help_me_carry.py
```diff
@@ -0,0 +1,188 @@
...
+
+    def __init__(self):
+        """Initialize the node"""
+        super().__init__("Help_me_carry_task_manager")
+        self.subtask_manager = SubtaskManager(
+            self, task=Task.HELP_ME_CARRY, mock_areas=["manipulation", "navigation"]
+        )
+        self.current_state = HelpMeCarryTM.TASK_STATES["START"]
+        self.current_attempts = 0
```
IvanRomero03, line 42: unused??

task_manager/scripts/help_me_carry.py
```diff
@@ -0,0 +1,188 @@
...
+        self.subtask_manager.manipulation.move_to_position("navigation")
+        future = self.subtask_manager.nav.move_to_location(location, sublocation)
+        rclpy.spin_until_future_complete(self, future)
+
+    def confirm(self, statement: str) -> bool:
+        """Confirm the name is correct"""
+        self.subtask_manager.hri.say(f"I heard {statement}, is that correct?")
+        response = self.subtask_manager.hri.hear()
+        return True
+        return self.subtask_manager.hri.is_positive(response)
```
IvanRomero03, line 60:
```suggestion
        return self.subtask_manager.hri.is_positive(response)
```

task_manager/scripts/help_me_carry.py
```diff
@@ -0,0 +1,188 @@
...
+                self.subtask_manager.hri.say("Please stand in front of me")
+
+        if self.current_state == HelpMeCarryTM.TASK_STATES["CONFIRM_FOLLOWING"]:
+            Logger.state(self, "Confirming following to guest")
+            self.subtask_manager.hri.say("I will follow you now")
+            self.current_state = HelpMeCarryTM.TASK_STATES["FOLLOWING_TO_DESTINATION"]
+
+        if self.current_state == HelpMeCarryTM.TASK_STATES["FOLLOWING_TO_DESTINATION"]:
+            Logger.state(self, "Following to destination")
+            while not self.subtask_manager.hri.hear("STOP"):
```
IvanRomero03, line 101: can we have a timeout here?

task_manager/scripts/help_me_carry.py
```diff
@@ -0,0 +1,188 @@
...
+                # TODO: person_pose = self.subtask_manager.vision.follow_person() /It must return the person pose and the person position towards the center of the camera
```
Ale-Coeto, line 102: For now the follow person is triggered with:
```python
self.subtask_manager.vision.track_person()
```
This sets the target to the largest person in the image and begins publishing the results for nav to use. Should it also return the current point when the service is called? Or should manipulation subscribe to the topic?

I would say it could be better for each area to subscribe to the topic independently

Ale-Coeto: Also, there is now a service that tracks the person who is raising the hand. It could perhaps be used for a deux ex machina
(not yet in subtask manager, but will be ready tomorrow (thursday))

Ale-Coeto (approved):
LGTM

task_manager/scripts/help_me_carry.py
Oscar-gg, line 60: The most recent update of hri's subtask manager implements the [confirm](https://github.com/RoBorregos/home2/blob/main/task_manager/scripts/subtask_managers/hri_tasks.py#L212) method. I would recommend to use it to handle confirmations. Please let me know if there is an additional argument/behaviour you would like the `confirm` to have.

task_manager/scripts/help_me_carry.py
```diff
@@ -0,0 +1,188 @@
...
+    def confirm(self, statement: str) -> bool:
+        """Confirm the name is correct"""
+        self.subtask_manager.hri.say(f"I heard {statement}, is that correct?")
+        response = self.subtask_manager.hri.hear()
+        return True
+        return self.subtask_manager.hri.is_positive(response)
+
+    def hear_word(self, word: str) -> bool:
+        """Check if the word is heard"""
+        statement = self.subtask_manager.hri.hear()
```
Oscar-gg, line 64: the new hri subtask also modifies the output of all functions to `status, res = hri_method(...)` or `status = hri_method(...)` if the original function was void. I think these implies changes for most of the calls to the hri subtask

PR #252, opened 2025-04-10
Vision subtask

Ale-Coeto:
Update vision subtask manager to handle new moondream setup
Update run and launch scripts for carry_my_luggage
Fix run script to not re-build container if already running
Added bag description to subtask manager


EmilianoHFlores (approved):
very good

PR #253, opened 2025-04-12
Working pick with object detection

EmilianoHFlores:
Includes

- ZED and gripper added to simulation with option load_zed:=true
- Fixed object detector transform issues
- Refactored pick and place code -> see PickTask.py at pick_and_place package
- Included object detection handler service
- Added manual input to test and trigger pick and place tasks using keyboard inputs -> see keyboard_input.py on pick_and_place package

Tested on simulation and real robot


Deivideich (approved):
Lgtm :+1:

PR #258, opened 2025-04-12
Pointinig object service

EmilianoHFlores:
Includes

- A service listening to zero shot detection, returning  the object the user is pointing at when called
- Usage on subtask manager and task manager example


Ale-Coeto (approved):
LGTM

PR #259, opened 2025-04-14
GPSR - pose or gesture one person

DanaeSG:
- Returns the pose or gesture of the person with the biggest bbox


vision/packages/moondream_run/scripts/moondream_node.py
Ale-Coeto, line 1: good catch

Ale-Coeto (commented):
Looks good, only check the ruff and moondream.
Also, maybe in a different pr, but we'll need to add it to the vision_tasks

Ale-Coeto (commented):
Nicee, just ensure ruff passes and there are no merge conflicts

PR #277, opened 2025-04-17
Integration follow person node

GerardoFJ:
Carry my lougage final commits


Deivideich (approved):
It works well, even thought, the control tuning of the arm rotation can be improved to have a smoother behavior

PR #283, opened 2025-04-21
Hri display localhost

GilMM27:
* Don't print console.logs
* Use build when running
* localhost works on orin


Oscar-gg (approved):
LGTM 🚀

PR #293, opened 2025-04-23
Add commands db

lleo-13:
added the methods 
-add command history
-query command history
-find closest

### DIFF
--- a/hri/packages/embeddings/scripts/ChromaAdapter.py
+++ b/hri/packages/embeddings/scripts/ChromaAdapter.py
@@ -86,6 +86,7 @@ class ChromaAdapter:
         - Normalizes metadatas to a list of dictionaries.
         - Cleans metadata with remove_empty_lists and remove_nulls.
         """
+
         collection_ = self.get_collection(collection_name)
 
         # Ensure documents is always a list
--- a/hri/packages/embeddings/scripts/ChromaAdapter.py
+++ b/hri/packages/embeddings/scripts/ChromaAdapter.py
@@ -16,6 +16,13 @@ class ChromaAdapter:
             )
         )
 
+    def delete_collection(self, collection_name: str):
+        """Method to delete a collection"""
+        try:
+            self.client.delete_collection(name=collection_name)
+        except Exception:
+            raise ValueError(f"The collection is missing {collection_name}")
+
     def remove_all_collections(self):
         """Method to remove all collections"""
         collections = self.client.list_collections()
--- a/task_manager/scripts/gpsr_task_manager.py
+++ b/task_manager/scripts/gpsr_task_manager.py
@@ -11,6 +11,7 @@ from subtask_managers.gpsr_tasks import GPSRTask
 from subtask_managers.gpsr_test_commands import get_gpsr_comands
 from utils.logger import Logger
 from utils.status import Status
+from subtask_managers.hri_tasks import HRITasks
 from utils.subtask_manager import SubtaskManager, Task
 
 ATTEMPT_LIMIT = 3
@@ -44,6 +45,7 @@ class GPSRTM(Node):
     def __init__(self):
         """Initialize the node"""
         super().__init__("gpsr_task_manager")
+        self.hri_tasks = HRITasks(self.subtask_manager)
         self.subtask_manager = SubtaskManager(self, task=Task.GPSR, mock_areas=["navigation"])
         self.gpsr_tasks = GPSRTask(self.subtask_manager)
         self.gpsr_individual_tasks = GPSRSingleTask(self.subtask_manager)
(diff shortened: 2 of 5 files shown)


task_manager/scripts/gpsr_task_manager.py
```diff
@@ -107,7 +109,14 @@ def run(self):
                 else:
                     Logger.info(self, f"Executing command: {command}")
                     self.subtask_manager.hri.say(f"Executing command: {command}")
-                    exec_commad(command["complement"], command["characteristic"])
+                    status, res = exec_commad(command["complement"], command["characteristic"])
```
Oscar-gg, line 112: Nice!

task_manager/scripts/subtask_managers/hri_tasks.py
```diff
@@ -404,38 +405,217 @@ def is_positive(self, text):
         future = self.is_positive_service.call_async(request)
         rclpy.spin_until_future_complete(self.node, future)
         return Status.EXECUTION_SUCCESS, future.result().is_positive
-
-    @service_check("is_negative_service", (Status.SERVICE_CHECK, False), TIMEOUT)
+        @service_check("is_negative_service", (Status.SERVICE_CHECK, False), TIMEOUT)
```
Oscar-gg: I think the indentation of this wrapper has an extra tab, please remove the extra tab

task_manager/scripts/subtask_managers/hri_tasks.py
```diff
@@ -404,38 +405,217 @@ def is_positive(self, text):
...
+                "status": status,
+                "timestamp": datetime.now().isoformat(),
+            }
+        ]
+        self.node.get_logger().info(f"Saving command history: {document}, {metadata}")
+
+        # Prepare the service request
+        metadata = [json.dumps(metadata)]
+
+        request = AddEntry.Request(document=document, metadata=metadata[0], collection=collection)
```
Oscar-gg, line 437: If the array isn't needed, consider passing directly to reduce the loc

For example:
```python3
request = AddEntry.Request(document=document, metadata=json.dumps(metadata), collection=collection)
```

which I think should be equivalent

hri/packages/embeddings/scripts/ChromaAdapter.py
Oscar-gg, line 1: I hadn't realized, but I think line 140 is duplicated

task_manager/scripts/subtask_managers/hri_tasks.py
```diff
@@ -404,38 +405,217 @@ def is_positive(self, text):
...
+    def add_location(self, document: list, metadata: str) -> list[str]:
         return self._add_to_collection(document, metadata, "locations")
 
+    def query_item(self, query: str, top_k: int = 1) -> list[str]:
+        return self._query_(query, "items", top_k)
+
+    def query_location(self, query: str, top_k: int = 1) -> list[str]:
+        return self._query_(query, "locations", top_k)
+
+    def find_closest(self, documents: list, query: str, top_k: int = 1) -> list[str]:
```
Oscar-gg, line 462: For this function would it be possible to have a flag to specify that the comparison should only be performed with the documents and query passed to the function in a single call? From my understanding, the current function compares the given query with all the inserted elements across all calls.

I think that having independent calls could have some use cases. For instance, when we need to match against items that are initially unknown.

task_manager/scripts/subtask_managers/hri_tasks.py
```diff
@@ -404,38 +405,217 @@ def is_positive(self, text):
...
+        Extracts the 'context' from the metadata of a query result.
+
+        Args:
+            query_result (tuple): The query result tuple (status, list of JSON strings)
+
+        Returns:
+            str: The 'context' field from metadata, or empty string if not found
+        """
+        try:
+            parsed_result = json.loads(query_result[1][0])  # parse the first JSON string
```
Oscar-gg, line 519: I think we could refactor all the getters as there is significant overlap in the implementation.

For example, we could create an extract metadata method:
```python3
def get_metadata_key(self, metadata_field, query_result):
        """
        """
        try:
            parsed_result = json.loads(query_result[1][0])  # parse the first JSON string
            metadata = parsed_result["results"][0]["metadata"]  # go into metadata
            context = metadata.get(metadata_field, "") 
            return context
        except (IndexError, KeyError, json.JSONDecodeError) as e:
            self.get_logger().error(f"Failed to extract context: {str(e)}")
            return ""
``` 

And then reuse it like this:
```python3
def get_context(self, query_result)
    return get_metadata_key('context', query_result)
def get_complement(self, query_result)
    return get_metadata_key('complement', query_result)
... # And so on
```

Oscar-gg (changes requested):

task_manager/scripts/subtask_managers/hri_tasks.py
lleo-13: solved

hri/packages/embeddings/scripts/ChromaAdapter.py
lleo-13: Is intended because each element in the list is a stringified JSON string inside a stringified list.

task_manager/scripts/subtask_managers/hri_tasks.py
lleo-13: solved

Oscar-gg (approved):
lgtm, nice work!!

PR #300, opened 2025-04-24
Hri debug speech

Oscar-gg:
Save inferences with audio


hri/packages/speech/speech/wav_utils.py
```diff
@@ -1,9 +1,10 @@
+import os
```
GilMM27, line 1: why did import order change here?

hri/packages/speech/scripts/stt/faster-whisper.py
```diff
@@ -11,20 +12,35 @@
...
+            os.makedirs(self.log_dir, exist_ok=True)
 
     def load_model(self):
         model_directory = os.path.join(os.path.dirname(__file__), "models")
+        device = "cpu"
+        try:
+            import torch
+
+            if torch.cuda.is_available():
+                device = "cuda"
```
GilMM27, line 36: I'd like if it printed when cuda was being used, just to make sure

GilMM27 (approved):
lgtm, thanks for deleting the other faster-whisper script. Added some suggestions

hri/packages/speech/speech/wav_utils.py
Oscar-gg: Ruff stuff

Oscar-gg: I think

PR #301, opened 2025-04-24
feat: added cache implementation

Oscar-gg:
Change to speed up speech pipeline
- Cache last 100 audios


GilMM27 (approved):
this looks great!

GilMM27:
what happens when the limit audio amount is reached?

PR #303, opened 2025-04-24
Improve display

GilMM27:
* Open automatically when display is ready
* Start web video server on compose up
* Display video from topic
* Change topic with by topic pub
* Improved visibility of vad detector


Oscar-gg (approved):
LGTM!!!!!

PR #308, opened 2025-04-25
Hri display video refresh

GilMM27:
* Shows video when it becomes available
* Changing the topic changes the video feed


Oscar-gg (approved):
🐐

PR #314, opened 2025-04-26
Hri stt filter

GilMM27:
* Added model warmup
* filter audio segment based on dynamic decibel threshold


docker/hri/stt-l4t.yaml
Oscar-gg, line 1: :rocket:

Oscar-gg (approved):
LGTM  :rock:

PR #321, opened 2025-04-27
separated build

GilMM27:


Oscar-gg (approved):
nice lgtm. maybe specify in a readme that the build of the display should be made

PR #323, opened 2025-04-27
LookAt Function

EmilianoHFlores:
Added a function to get the orientation so that the camera can look directly at a set point. Example usage at look_at.py on frida_motion_planning/examples


Ale-Coeto (approved):
LGTM

PR #325, opened 2025-04-28
display update

GilMM27:
* Don't show hear service mock keywords
* Publish transcription even on service calls
* Don't print web video server logs


docker/hri/display-entrypoint.sh
Oscar-gg, line 1: :pray:

Oscar-gg (approved):
lgtm :rocket:

PR #333, opened 2025-04-30
Implementation of Constrained Planning for Manipulation

richy-gs:


manipulation/packages/frida_motion_planning/frida_motion_planning/robots/xarm6.py
EmilianoHFlores, line 1: why did we move this here? it's fine by me if we remove the frida_pymoveit2 package tbh just be sure nothing is broken

manipulation/packages/frida_motion_planning/examples/ex_orientation_path_constraint.py
EmilianoHFlores, line 1: good

manipulation/packages/frida_motion_planning/examples/ex_orientation_path_constraint.py
```diff
@@ -0,0 +1,119 @@
...
+        goal_msg.acceleration = acceleration
+        goal_msg.planner_id = planner_id
+        goal_msg.apply_constraint = useConstraint
+
+        if useConstraint:
+            constraints_msg = Constraint()
+            constraints_msg.orientation = pose.pose.orientation
+            constraints_msg.frame_id = robot.base_link_name()
+            constraints_msg.target_link = robot.end_effector_name()
+            constraints_msg.tolerance_orientation = [3.14159, 0.5, 0.5]
```
EmilianoHFlores, line 54: can you add documentation on what you have synthesized on the parameters? like, are the three tolerances rpy? what the weight is and such. Just a one line description should do the trick, or point to a link that has the info

EmilianoHFlores (changes requested):

EmilianoHFlores:
I only need to be sure that the new xarm6.py doesnt break anything, all other stuff we can leave for a next pr

manipulation/packages/frida_motion_planning/frida_motion_planning/robots/xarm6.py
richy-gs: It was just for my test, but I tried now with the original file in frida_pymoveit2 and it worked fine, so I can delete the one I created. In MoveItPlanner, the file is used "from frida_pymoveit2.robots import xarm6", so it might be a good idea to delete my file to avoid confusion.

manipulation/packages/frida_motion_planning/examples/ex_orientation_path_constraint.py
richy-gs: Ok, I'll add some comments

EmilianoHFlores (approved):
great!

PR #336, opened 2025-05-01
feat: tmux scripting for receptionist

Oscar-gg:


scripting/receptionist/manipulation-planning.sh
Ale-Coeto, line 1: shouldnt this be the node to turn on the xarm? it has the same command as motion-planning

Oscar-gg: yes, fixed

scripting/receptionist/manipulation-follow-face.sh
Ale-Coeto, line 1: I think now there is a launch for manipulation

Ale-Coeto (approved):

PR #337, opened 2025-05-01
New Gripper Pick and Place

EmilianoHFlores:
Integrated new gripper and fixed several issues on pipeline:

- Now tries aletrnatives on pick_server to try closer/farther from the object
- Now tries first a horizontal pick then vertical
- Includes a quick feasibility test based on pick pose height to reject poses without trying plan
- Fixed a big, previously unnoticed issue where gripper was rotated 45 degrees from where GPD tried to send it
- Tested pick, upgraded params for stability
- Tested place and place on shelves, fully working and stable


IvanRomero03 (approved):
pap

PR #341, opened 2025-05-01
added individual models

lleo-13:
added config for individual models


Oscar-gg (changes requested):
Looks good. Please add it also for the data extractor node

Oscar-gg (approved):
LGTM

PR #361, opened 2025-05-05
Integrated new command interpreter call in GPSR task manager through BAML

afr2903:
**‼️ Actual test missing**

- The LLM inference is now directly in the subtask manager to avoid complex ROS msg transformation.
- Given that we lost generality (from previous standarized `action`,`complement`,`characteristic`) now the complete Pydantic object of each command is passed as param, to be managed internally in the method.
- The existent methods in `gpsr_tasks` and `gpsr_single_task` (someone explain to me the difference between files 😭) were upgraded to receive the command object as param.


Oscar-gg (approved):
LGTM

PR #363, opened 2025-05-05
Update Ollama image version in Docker configuration to support qwen3

Diego-HC:
*Requires manually building the ollama image with `jetson-containers build ollama`


Oscar-gg (approved):
LGTM :man_cook:

PR #370, opened 2025-05-08
refactor whereIam, returns the items inside the location ordered

lleo-13:
- Return areas service, now returns also the subareas inside organized by how far are they from the current location.
- Exceptions are when the location has no polygon to determine the area, and when the subarea is empty. In those cases it returns an empty string  and an empty array

### DIFF
--- a/task_manager/scripts/point_transformer.py
+++ b/task_manager/scripts/point_transformer.py
@@ -10,6 +10,7 @@ import os
 from ament_index_python.packages import get_package_share_directory
 from geometry_msgs.msg import TransformStamped
 from utils.status import Status
+from math import sqrt
 
 POINT_TRANSFORMER_TOPIC = "/integration/point_transformer"
 RETURN_AREAS_TOPIC = "/integration/return_areas"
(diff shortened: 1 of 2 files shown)


task_manager/scripts/point_transformer.py
```diff
@@ -88,50 +89,57 @@ def get_actual_pose(self, request, response):
...
+        """
+        Callback to determine the location of the robot based on its pose.
+        """
         package_share_directory = get_package_share_directory("frida_constants")
         file_path = os.path.join(package_share_directory, "map_areas/areas.json")
         mylocation = ""
+
+        # Load areas from the JSON file
         with open(file_path, "r") as file:
             self.areas = json.load(file)
```
Diego-HC: we don't need to have this as a class attribute

task_manager/scripts/point_transformer.py
```diff
@@ -146,6 +154,49 @@ def is_inside(self, x, y, polygon):
...
+                return response
+
+            response.nearest_locations = nearest_locations
+            response.location = mylocation
+            response.success = True
+            return response
+
+        except Exception as e:
+            self.get_logger().error(f"Could not get transform: {str(e)}")
+            response.location = None
```
Diego-HC: This should be an empty string, and you should also set response.nearest_locations to an empty list

Diego-HC (approved):
lgtm

PR #379, opened 2025-05-14
feat: faster is_positive, is_negative, and name and location extraction

Oscar-gg:
- Added spacy models to extract data
- Classifier for is_positive

TODO:
- Add testing files for is_positive and extract_data


afr2903 (approved):
The new approach for faster extraction is smart 🚀

PR #423, opened 2025-06-19
Adaptable poinCloud resolution

Fernando94654:
Add pointCloud with adaptable resolution depdending on distance to the robot
This changes can reduce the number of total points and increase the resolution in important areas optimizing subsequent processes, this is the result:
oll_resolution
-Old resolution (leaf size = 0.01) with +200,000 points
new_resolution
-New resollution (   small_size = 0.01, medium_size = 0.05, large_size = 0.10) with 17,000 points
better_resolution
-Better resolution ( small_size = 0.005, medium_size = 0.05, large_size = 0.10) with 51,000 points

### DIFF
--- a/frida_constants/include/frida_constants/manipulation_constants_cpp.hpp
+++ b/frida_constants/include/frida_constants/manipulation_constants_cpp.hpp
@@ -70,6 +70,4 @@ conststr DEBUG_POSE_GOAL_TOPIC = "/manipulation/debug_pose_goal";
 conststr GET_JOINT_SERVICE = "/manipulation/get_joints";
 conststr TOGGLE_SERVO_SERVICE = "/manipulation/toggle_servo";
 
-constexpr float SQ_SMALL_CLOUD_RADIUS = 1.5f * 1.5f;
-constexpr float SQ_MEDIUM_CLOUD_RADIUS = 2.5f * 2.5f;
 #endif // MANIPULATION_CONSTANTS_CPP
--- a/frida_constants/include/frida_constants/manipulation_constants_cpp.hpp
+++ b/frida_constants/include/frida_constants/manipulation_constants_cpp.hpp
@@ -70,4 +70,6 @@ conststr DEBUG_POSE_GOAL_TOPIC = "/manipulation/debug_pose_goal";
 conststr GET_JOINT_SERVICE = "/manipulation/get_joints";
 conststr TOGGLE_SERVO_SERVICE = "/manipulation/toggle_servo";
 
+constexpr float SQ_SMALL_CLOUD_RADIUS = 1.5f * 1.5f;
+constexpr float SQ_MEDIUM_CLOUD_RADIUS = 2.5f * 2.5f;
 #endif // MANIPULATION_CONSTANTS_CPP
--- a/manipulation/packages/perception_3d/src/down_sample_pc.cpp
+++ b/manipulation/packages/perception_3d/src/down_sample_pc.cpp
@@ -39,6 +39,8 @@ private:
   float small_size = 0.01f;
   float medium_size = 0.05f;
   float large_size = 0.10f;
+  const float SMALL_CLOUD_RADIUS = 1.5f * 1.5f;
+  const float MEDIUM_CLOUD_RADIUS = 2.5f * 2.5f;
 
 public:
   DownSamplePointCloud() : Node("downsample_pointcloud") {
@@ -76,8 +78,8 @@ public:
     for(size_t i=0; i < in_cloud->points.size(); i++){
       const auto& pt = in_cloud->points[i];
       float sq_radius = (pt.x * pt.x + pt.y * pt.y + pt.z * pt.z);
-      if(sq_radius < SQ_SMALL_CLOUD_RADIUS) small_cloud->points.push_back(pt); 
-      else if(sq_radius < SQ_MEDIUM_CLOUD_RADIUS) medium_cloud->points.push_back(pt);
+      if(sq_radius < SMALL_CLOUD_RADIUS * SMALL_CLOUD_RADIUS) small_cloud->points.push_back(pt); 
+      else if(sq_radius < MEDIUM_CLOUD_RADIUS * MEDIUM_CLOUD_RADIUS) medium_cloud->points.push_back(pt);
       else large_cloud->points.push_back(pt);
     }
     //insert clouds
--- a/manipulation/packages/perception_3d/src/down_sample_pc.cpp
+++ b/manipulation/packages/perception_3d/src/down_sample_pc.cpp
@@ -41,7 +41,7 @@ private:
   float large_size = 0.10f;
   float small_radius = 1.5f; // 1.5m
   float medium_radius = 2.5f; // 2.5m
-  float sqr_small_rad = std::pow(small_radius, 2); // x^2 + y^2+ z^2
+  float sqr_small_rad = std::pow(small_radius, 2);
   float sqr_med_rad = std::pow(medium_radius, 2);
 
 public:
--- a/manipulation/packages/perception_3d/src/down_sample_pc.cpp
+++ b/manipulation/packages/perception_3d/src/down_sample_pc.cpp
@@ -39,8 +39,10 @@ private:
   float small_size = 0.01f;
   float medium_size = 0.05f;
   float large_size = 0.10f;
-  const float SMALL_CLOUD_RADIUS = 1.5f;
-  const float MEDIUM_CLOUD_RADIUS = 2.5f;
+  float small_radius = 1.5f; // 1.5m
+  float medium_radius = 2.5f; // 2.5m
+  float sqr_small_rad = std::pow(small_radius, 2); // x^2 + y^2+ z^2
+  float sqr_med_rad = std::pow(medium_radius, 2);
 
 public:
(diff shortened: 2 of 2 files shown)


manipulation/packages/perception_3d/src/down_sample_pc.cpp
```diff
@@ -34,7 +36,12 @@ class DownSamplePointCloud : public rclcpp::Node {
 
   std::string input_topic = ZED_POINT_CLOUD_TOPIC;
   std::string output_topic = POINT_CLOUD_TOPIC;
-  float leaf_size = 0.01f;
+  float small_size = 0.01f;
+  float medium_size = 0.05f;
+  float large_size = 0.10f;
+
+  float small_cloud_radius = 1.5;
+  float medium_cloud_radius  = 2.5;
```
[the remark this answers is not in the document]
IvanRomero03: shouldnt be constants
@Fernando94654 please add them to the node params

manipulation/packages/perception_3d/src/down_sample_pc.cpp
```diff
@@ -59,14 +68,40 @@ class DownSamplePointCloud : public rclcpp::Node {
...
+      if(radius < small_cloud_radius) small_cloud->points.push_back(in_cloud->points[i]); 
+      else if(radius >= small_cloud_radius && radius < medium_cloud_radius) medium_cloud->points.push_back(in_cloud->points[i]);
+      else if(radius >= medium_cloud_radius) large_cloud->points.push_back(in_cloud->points[i]);
+    }
+    //insert small cloud
+    pcl::VoxelGrid<pointCloudType> small_sor;
+    small_sor.setInputCloud(small_cloud);
+    small_sor.setLeafSize(small_size, small_size, small_size);
+    small_sor.filter(*temporal_cloud);
+    sampled_cloud->insert(sampled_cloud->end(), temporal_cloud->begin(), temporal_cloud->end());
```
IvanRomero03, line 90: can you put this in a different function so we dont repeat ourselves here three times?

manipulation/packages/perception_3d/src/down_sample_pc.cpp
```diff
@@ -59,14 +68,40 @@ class DownSamplePointCloud : public rclcpp::Node {
...
     PointCloudNS::Ptr sampled_cloud(new PointCloudNS);
 
     pcl::fromROSMsg(*msg, *in_cloud);
 
-    pcl::VoxelGrid<pointCloudType> sor;
-    sor.setInputCloud(in_cloud);
-    sor.setLeafSize(leaf_size, leaf_size, leaf_size);
-    sor.filter(*sampled_cloud);
+    for(size_t i=0; i < in_cloud->points.size(); i++){
+      float radius = std::sqrt(std::pow(in_cloud->points[i].x, 2) + std::pow(in_cloud->points[i].y, 2) + std::pow(in_cloud->points[i].z, 2));
```
IvanRomero03, line 80: for optimization purposes what about not doing the srqt and comparing via squared distances?

manipulation/packages/perception_3d/src/down_sample_pc.cpp
```diff
@@ -59,14 +68,40 @@ class DownSamplePointCloud : public rclcpp::Node {
 
   void pointcloud_callback(const sensor_msgs::msg::PointCloud2::SharedPtr msg) {
     PointCloudNS::Ptr in_cloud(new PointCloudNS);
+    PointCloudNS::Ptr small_cloud(new PointCloudNS);
+    PointCloudNS::Ptr medium_cloud(new PointCloudNS);
+    PointCloudNS::Ptr large_cloud(new PointCloudNS);
+    PointCloudNS::Ptr temporal_cloud(new PointCloudNS);
```
IvanRomero03, line 74: a lot of cloud ptrs (write & read & write) is this the best way to accomplish this?

IvanRomero03 (changes requested):
- Please review the comments
- Please attach the description of why are we working towards this approach and a video/picture of the result.
- Also can we add here in the pr size comparison between the previous approach (or just setting all leaf sizes to 0.01f) and the current one? (Mainly to show why doing this will give more accurate results while reducing the total size)

manipulation/packages/perception_3d/src/down_sample_pc.cpp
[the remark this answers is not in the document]
Fernando94654: this params are now in manipulation_constants

manipulation/packages/perception_3d/src/down_sample_pc.cpp
Fernando94654: dedicate function "insert_cloud" was created

manipulation/packages/perception_3d/src/down_sample_pc.cpp
Fernando94654: squared distances were implemented

frida_constants/include/frida_constants/manipulation_constants_cpp.hpp
```diff
@@ -70,4 +70,6 @@ conststr DEBUG_POSE_GOAL_TOPIC = "/manipulation/debug_pose_goal";
 conststr GET_JOINT_SERVICE = "/manipulation/get_joints";
 conststr TOGGLE_SERVO_SERVICE = "/manipulation/toggle_servo";
 
+constexpr float SQ_SMALL_CLOUD_RADIUS = 1.5f * 1.5f;
+constexpr float SQ_MEDIUM_CLOUD_RADIUS = 2.5f * 2.5f;
```
IvanRomero03, line 74: please move these back to the down sample file and make them params of the node

IvanRomero03: And keep them as it radius value, you could later make a constant of the squared of the radius that holds its value. but its important to keep the semantic value here

IvanRomero03 (changes requested):

manipulation/packages/perception_3d/src/down_sample_pc.cpp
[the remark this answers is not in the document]
Fernando94654: @IvanRomero03 is it correct now?

manipulation/packages/perception_3d/src/down_sample_pc.cpp
Fernando94654: i think in something like that before, using only one additional cloud but would have to traverse the in_cloud 3 times making it slower. do you have any other idea??
   
    void filter2_cloud(const PointCloudNS::Ptr& input_cloud, PointCloudNS::Ptr& output_cloud, float& sq_min_radius, float& sq_max_radius, float leaf_size) {
      PointCloudNS::Ptr temporal_cloud(new PointCloudNS);
      for (const auto& pt : input_cloud->points) {
        float sq_radius = pt.x * pt.x + pt.y * pt.y + pt.z * pt.z;
        if (sq_radius > sq_min_radius &&  sq_radius < sq_max_radius) {
          temporal_cloud->points.push_back(pt);
        }
      }
      pcl::VoxelGrid<pointCloudType> sor;
      sor.setInputCloud(temporal_cloud);
      sor.setLeafSize(leaf_size, leaf_size, leaf_size);
      sor.filter(*temporal_cloud);
      output_cloud->insert(output_cloud->end(), temporal_cloud->begin(), temporal_cloud->end());
    }

manipulation/packages/perception_3d/src/down_sample_pc.cpp
```diff
@@ -34,7 +36,11 @@ class DownSamplePointCloud : public rclcpp::Node {
 
   std::string input_topic = ZED_POINT_CLOUD_TOPIC;
   std::string output_topic = POINT_CLOUD_TOPIC;
-  float leaf_size = 0.01f;
+  float small_size = 0.01f;
+  float medium_size = 0.05f;
+  float large_size = 0.10f;
+  const float SMALL_CLOUD_RADIUS = 1.5f * 1.5f;
+  const float MEDIUM_CLOUD_RADIUS = 2.5f * 2.5f;
```
IvanRomero03:
```suggestion
  float small_radius = 1.5f // 1.5m
  float medium_radius = 2.5f // 2.5m
  float sqr_small_rad = std::pow(small_radius, 2)*3; // x^2 + y^2+ z^2
  float sqr_med_rad = std::pow(medium_radius, 2)*3;
  
  ...
  this->small_radius = this->declare_parameter( ...
   float sqr_small_rad =  ...
```

manipulation/packages/perception_3d/src/down_sample_pc.cpp
IvanRomero03: voxelgrid was a nice approach for this one in our first usecase and an easy way to tackle this one, I was thinking maybe implementing this one differently maybe random sampling based on distance or something like that (or implementing something similar as VoxelGrid but for us, but maybe for another time, for this pr lets stick to voxelgrid.

manipulation/packages/perception_3d/src/down_sample_pc.cpp
Fernando94654: change made, but the rad should not be multiplied by 3 to obtain sq_rad

IvanRomero03 (approved):
Very nice work!

PR #425, opened 2025-06-20
check if person is inside the house

DanaeSG:


vision/packages/vision_general/scripts/is_person_inside.py
```diff
@@ -0,0 +1,196 @@
...
+class Person3DEstimator:
+    def __init__(self, camera_frame=CAMERA_FRAME):
+        self.pose_detector = PoseDetection()
+        self.camera_frame = camera_frame
+
+    def estimate(self, image_bgr, depth_image, camera_info):
+        image_rgb = cv2.cvtColor(image_bgr, cv2.COLOR_BGR2RGB)
+        results = self.pose_detector.detect(image_rgb)
+
+        if not results.pose_landmarks or len(results.pose_landmarks.landmark) <= 12:
```
GilMM27, line 42: will this fail if the person is only partially inside the feed and should this be the expected behavior?

vision/packages/vision_general/scripts/pose_detection.py
```diff
@@ -15,6 +14,9 @@ def __init__(self):
         self.pose = self.mp_pose.Pose()
         self.mp_drawing = mp.solutions.drawing_utils
 
+    def detect(self, image_rgb):
```
GilMM27, line 17: does this replace `personAngle` or why are we deleting it? Wasn't that function used for reID?

vision/packages/vision_general/scripts/gpsr_commands.py
```diff
@@ -233,6 +237,13 @@ def count_gestures(self, frame):
             # Crop the frame to the bounding box of the person
             cropped_frame = frame[y1:y2, x1:x2]
 
+            is_inside = self.person_inside_client.call(y1, x1, y2, x2)
```
GilMM27, line 240: I'd rather we'd be a bit more descriptive like -> `person_in_sight` or `person_inside_video`

PR #426, opened 2025-06-20
update image in ollama-jetson compose file

Diego-HC:
use docker hub image to avoid having to manually build the image


Oscar-gg (approved):
LGTM!

PR #442, opened 2025-06-25
read QR service implemented in gpsr_commands

Fernando94654:
In gpsr the commands can be given through a qr code, so in this pr a service for this was implemented using cv2.QRCodeDetector ans return a string value with the command
The following video shows a test using zed_simulator and client tester
[read_qr_test.webm](https://github.com/user-attachments/assets/2a9bc050-df63-470e-aa0a-4337e6c0723e)
and this is the test_client.py, if you want to run, add this script to home2/vision/packages/vision_general/scripts/:
```
#!/usr/bin/env python3
import sys
import cv2
from frida_interfaces.srv import ReadQr
from frida_constants.vision_constants import READ_QR_TOPIC
import rclpy
from rclpy.node import Node

class client(Node):
    def __init__(self):
        super().__init__('test_client')
        self.client = self.create_client(ReadQr, READ_QR_TOPIC)
        self.req = ReadQr.Request()

    def send_request(self):
        self.req.request = True
        return self.client.call_async(self.req)

def main():
    rclpy.init()
    t_client = client()
    future = t_client.send_request()
    rclpy.spin_until_future_complete(t_client, future)
    response = future.result()
    print("El QR dice")
    print(response.result)

if __name__ == "__main__":
    main()

### DIFF
--- a/vision/packages/vision_general/scripts/gpsr_commands.py
+++ b/vision/packages/vision_general/scripts/gpsr_commands.py
@@ -382,6 +382,13 @@ class GPSRCommands(Node):
         # Detect QR using cv2.QRCodeDetector
         retval, _, _ = self.qr_detector.detectAndDecode(frame)
 
+        if retval == "":
+            response.success = False
+            response.result = ""
+            self.get_logger().warn("No qr code detected")
+            return response
+
+        self.success(f"QR code read successfully: {retval}")
         response.result = retval
         response.success = True
         return response

### CODE REVIEW

File: vision/packages/vision_general/scripts/gpsr_commands.py [hunk 1]
```diff
@@ -358,6 +368,24 @@ def detect_pose_gesture_callback(self, request, response):
...
+            return response
+
+        frame = self.image
+        self.output_image = frame.copy()
+
+        # Detect QR using cv2.QRCodeDetector
+        retval, _, _ = self.qr_detector.detectAndDecode(frame)
+
+        response.result = retval
+        response.success = True
```
Ale-Coeto, line 386: I would suggest adding a success message or warning if there was no detection. (This is very useful for debugging a full task.

```python
# Success
self.success(f"QR code read successfully: {message})

# Warning
self.get_logger().warn("No qr code detected")
```

Ale-Coeto: Also it could be useful to check if the response was successful. For example if there is no qr code detected or if the message is empty it would be good to know so that the `task manager` could respond to that case and potentially re-try.

Ale-Coeto (commented):
Nice, just added a few comments

vision/packages/vision_general/scripts/gpsr_commands.py
Fernando94654: sucess and warning messages were added

Ale-Coeto (approved):
Nice! LGTM

PR #445, opened 2025-06-26
remove chroma adapter from unused place

IvanRomero03:
chroma adapter wasnt used here, moved away in order to reformat it


Oscar-gg (approved):
Lgtm

PR #450, opened 2025-06-28
Whisper live

Oscar-gg:
Whisper live implementation
- Action server implemented to access transcription and obtain feedback
- Hear method now uses `hear_streaming` to handle calls
- Keyword interpretation now also considers the text interpreted with `hear_streaming` in addition to using OWW.
- Bidirectional communication implemented using gRPC to obtain feedback from whisper live.


Diego-HC (approved):
LGTM! :man_cook:

PR #461, opened 2025-07-03
ReID Tracker Batched Operations

EmilianoHFlores:
Includes:

- Batch image feature extraction and comparison
- Now continuously extracts features to update person's info
- Now uses torch for all matrix operations, with cuda if enabled


IvanRomero03 (approved):
lgtm

PR #463, opened 2025-07-04
add layer for caching frida interfaces and constants

Diego-HC:
add docker compose for building frida_interfaces and frida_constants, and update integration docker setup to use the new layer.

To use, run `docker compose up` to build `frida_interfaces` and `frida_constants`. The resulting files will be automatically used in the integration container. The other docker setups will also be updated to support this cache in the future.


Oscar-gg (approved):
LGTM 🐐

PR #464, opened 2025-07-04
Postgresql locations

Oscar-gg:
- Add locations from `frida_constants`
- Modified to use cosine similarity in operations.


IvanRomero03 (approved):
goat

PR #516, opened 2025-07-07
Hand display

### LINKED ISSUE #515: Display disambiguate duplicated target locations
Idea: in give me a hand, the user can request to place the object relative to another object. However, there can be several objects of the same type. The idea is to display all possibilities to the user such that the user can say which one it is referring to. (i.e., show each location as a color and tell the user to say the color of the desired location)

Oscar-gg:
Closes #515 

image


CamilaTite26 (approved):
lgtm nice implementation

PR #517, opened 2025-07-07
EGPSR Task Manager

DanaeSG:
**Flow Overview:**
1. **Initialization**: Robot starts at start_area, introduces itself, and plans optimal exploration route using Dijkstra's algorithm
2. **Exploration Phase**: Systematically visits all areas (living_room, kitchen, bedroom, dining_room) to detect:
   - Trash objects (limited to 1 detection via flag)
   - Misplaced objects (TODO: HRI verification)
   - People with raised arms (left/right arm detection) (I remember it assumes the person is in front of the robot, so verify if this works)
3. **Cleanup Phase**: 
   - Handles trash: asks person to hand over trash → navigates to nearest trashbin → disposes trash (approach: current location, next one if current doesn't have a trash bin or previous one if current is the last location)
   - Handles misplaced objects: picks up → navigates to correct location → places object 
4. **Command Execution**: Prioritizes people with raised arms, asks for commands, interprets via LLM, and executes using GPSR subtasks
5. **Completion**: Returns to start position after MAX_COMMANDS (3) or task completion

**States**: START → EXPLORATION → HANDLING_TRASH → NAVIGATE_TO_TRASH_BIN → DROP_TRASH → PICK_MISPLACED_OBJECT  → PLACE_MISPLACED_OBJECT  → EXPLORATION →    WAITING_FOR_COMMAND → EXECUTING_COMMAND → FINISHED_COMMAND → DONE


task_manager/scripts/egpsr_task_manager.py
```diff
@@ -0,0 +1,579 @@
...
+        self.current_attempt = 0
+        self.executed_commands = 0
+        self.commands = []
+        self.exploration_locations = []
+        self.current_exploration_index = 0
+        self.found_misplaced_objects = []
+        self.found_trash = []
+        self.people_to_ask_command = []
+        self.exploration_complete = False
+        self.trash_detected_flag = False
```
Oscar-gg, line 238: Maybe we can use a counter instead of the flag.

For this task the optimal solution would be to solve one of each type of problem to avoid point deductions. However, I'm unsure if our robot can actually do that 😢 

So maybe if we leave it as a counter we could modify it faster to prioritize one of the specific problems.

task_manager/scripts/egpsr_task_manager.py
```diff
@@ -0,0 +1,579 @@
...
+            named_position="front_stare", velocity=0.5, degrees=True
+        )
+
+        # Detect objects and analyze them
+        self.analyze_location_for_anomalies(current_location)
+
+        # Move to next location
+        self.current_exploration_index += 1
+
+    def analyze_location_for_anomalies(self, location: str):
```
Oscar-gg, line 318: I like the approach of recording the problems and then determine if we should solve them, for instance using the `self.people_to_ask_command`.

I think we could also use this for the other types of problems so that after finding all the problems we can choose which ones to solve, and also have a flag to determine if we should solve them right away.

What do you think?

Oscar-gg (approved):
LGTM, very nice!!

Added some comments 👀

task_manager/scripts/egpsr_task_manager.py
```diff
@@ -0,0 +1,579 @@
...
+        Logger.info(self, f"Planned exploration order: {self.exploration_locations}")
+        Logger.info(self, f"Total exploration distance: {total_distance:.2f} meters")
+
+        # Load areas from the JSON file
+        with open(file_path, "r") as file:
+            areas = json.load(file)
+
+        # Check which area the robot is in
+        for area in areas:
+            self.trash_bin_location = areas.get(area, {}).get("trashbin", [])
```
afr2903, line 263: Is there only one trashbin in the house?
Otherwise, I think there's a bug here, isn't `trash_bin_location` being overwritten at each for loop iteration?

Something like this would store the trashbins in all rooms:
```python
self.trash_bin_locations = {}
for name, data in areas.items():
    if "trashbin" in data:
        self.trash_bin_locations[name] = data["trashbin"]
```

task_manager/scripts/egpsr_task_manager.py
afr2903, line 318: I believe the `analyze_location_for_anomalies` method has a lot of responsibilities, other than just analyzing anomalies. This could make debugging more complicated.

I'd add submethods to handle pieces of the implementation (e.g. `detect_trash()`).
And delegate the state machine handling in this function, to the `run` method.

afr2903 (commented):
🤯

task_manager/scripts/egpsr_task_manager.py
```diff
@@ -0,0 +1,579 @@
...
+    def __init__(self, areas_json_path: str):
+        """Initialize the exploration planner with areas data"""
+        with open(areas_json_path, "r") as file:
+            self.areas = json.load(file)
+
+        # Extract valid exploration areas (exclude start_area and entrance)
+        self.exploration_areas = {
+            name: data
+            for name, data in self.areas.items()
+            if name not in ["start_area", "entrance"]
```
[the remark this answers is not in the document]
DanaeSG: We are skipping main areas and polygon is a subarea. As long as we don't call it explicitly, polygons are not taken into account

task_manager/scripts/egpsr_task_manager.py
DanaeSG: I have divided the function into sub-methods for better maintainability.

task_manager/scripts/egpsr_task_manager.py
DanaeSG: Added solved by type counter and max solved constants

task_manager/scripts/egpsr_task_manager.py
DanaeSG: added a flag "solve immediately" to test different strategies. Also, now the locations of trash or misplaced objects are saved to solve the optimal problems at the end of the exploration.

PR #527, opened 2025-07-09
Update to not load pretrained swin model

Ale-Coeto:
```python
timm.create_model("swin_base_patch4_window7_224", pretrained=True)
```
This line was loading a pretrained model from the internet. Now it loads only the structure and later on the weights. The load was also unnecessary in `reid_model.py`.

*Needs testing in FRIDA


Ale-Coeto:
> Can this pr be closed?

Yes

PR #564, opened 2025-09-06
Improved run.sh for hri

GilMM27:
* Added missing area `zed` to `auto-complete.sh`, improved implementation and added hri specific arguments.
* Run `npm run start` instead of `dev`.
* Implemented `--open-display`, `--build-display` and `--download-model` arguments in `docker/hri/run.sh`.

### DIFF
--- a/docker/hri/display.yaml
+++ b/docker/hri/display.yaml
@@ -13,9 +13,8 @@ services:
     environment:
       ROS_LOCALHOST_ONLY: 1
     ports:
-      - "3000:3000"
-      - "8001:8001"
-      - "8080:8080"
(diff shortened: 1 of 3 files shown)


docker/hri/run.sh
Oscar-gg, line 1: Was this tested in the Orin? (Just in case  lol)

docker/hri/display.yaml
Oscar-gg, line 1: Why is the entrypoint script not used?
(The one used for the display in the Orin)

Oscar-gg (approved):
Lgtm

PR #574, opened 2025-09-26
572 modularize displays components

CamilaTite26:
https://github.com/user-attachments/assets/9d07307b-51d7-4239-b4ae-82141ddde2e8

https://github.com/user-attachments/assets/061a1e07-0c8b-4a9d-9aa7-c83c553e6483

To modularize the display components, they were separated into more files.


hri/display/web-ui/app/components/index.ts
```diff
@@ -0,0 +1,16 @@
+// Main component exports
+export { default as RosMessagesDisplay } from "./ros-messages";
```
GilMM27, line 2: is this file necessary?

hri/display/web-ui/app/components/useWebSocket.ts
```diff
@@ -0,0 +1,73 @@
+"use client";
```
GilMM27, line 1: ditto since this isn't a component but a function

GilMM27 (changes requested):
small nits, nice work

hri/display/web-ui/app/components/index.ts
CamilaTite26: We could remove it, but we'd need to update import statements across the codebase and it would make future refactoring harder.

CamilaTite26:
https://github.com/user-attachments/assets/bb287955-3811-4979-be56-323466bc0933

Requested changes

PR #578, opened 2025-10-05
fix permission errors when downloading tts model to volume

GilMM27:
`HF_HUB_DISABLE_XET=1`

forces HTTP instead of the xet backend that is causing the permission error


Oscar-gg (approved):

Oscar-gg:
> For some weird reason, running the script directly in the docker compose provoked an error when downloading the tts model but running an `entrypoint.sh` works. Also creating the dir for audios avoid permission errors when saving audios.

I think we should try to undestand the reason to fix similar future issues

PR #594, opened 2025-10-21
Runsh rework 2

GilMM27:
Depends on #595
* `run.sh` checks if base image exists or automatically builds it
* Added `frida_constants` area to build cache dirs
* Also checks if they have never been built and automatically builds them

test in

- [x] cpu
- [ ] cuda
- [ ] l4t

### DIFF
--- a/docker/hri/compose/docker-compose-l4t.yml
+++ b/docker/hri/compose/docker-compose-l4t.yml
@@ -1,7 +1,7 @@
 include:
-  - path: l4t-devices.yaml
+  - path: hri-ros-l4t.yaml
   - path: stt-l4t.yaml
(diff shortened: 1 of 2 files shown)


docker/frida_interfaces_cache/docker-compose-cuda.yaml
```diff
@@ -0,0 +1,17 @@
+services:
```
Oscar-gg: Is there any dependency that uses cuda in the interfaces cache?

GilMM27: I'm not sure but I guess it makes sense since users only have to build one base image

PR #595, opened 2025-10-21
force kill hri dockers on double ctr + c

GilMM27:
previous method started failing and double ctr + c just killed the sh, some seconds later the containers would stop. This ensures it kills them


Oscar-gg (approved):
LGTM

Oscar-gg:
Q: is there a way of integrating this to the general `run.sh`? (so that this doesn't need to be individually implemented by each area-specific `run.sh`)

PR #617, opened 2025-12-01
Cleaned ollama container moving the model download to docker

luisrosales852:
Moved download model logic (the .sh) to docker folder, updated docker configuration and gitignore. should work


.gitignore
```diff
@@ -12,11 +12,11 @@
...
 **.gz
-hri/packages/nlp/assets/**
+docker/hri/scripts/ollama/**
 .cache
-!hri/packages/nlp/assets/download-model.sh
-!hri/packages/nlp/assets/Modelfile
-!hri/packages/nlp/assets/entrypoint.sh
+!docker/hri/scripts/ollama/download-model.sh
+!docker/hri/scripts/ollama/Modelfile
+!docker/hri/scripts/ollama/entrypoint.sh
```
GilMM27: this shouldn't be needed anymore since `/ollama` isn't ignored

.gitignore
```diff
@@ -12,11 +12,11 @@
 **.pt
 **.pth
 **.gz
-hri/packages/nlp/assets/**
+docker/hri/scripts/ollama/**
```
GilMM27: we don't want to ignore this

GilMM27 (changes requested):
please test and add the missing changes ;)

docker/hri/compose/ollama-cuda.yaml
```diff
@@ -10,10 +10,11 @@ services:
     network_mode: host
     volumes:
       - ../../../hri/packages/nlp/assets:/ollama
+      - ../scripts/ollama:/scripts
```
GilMM27: I'm not sure whats the better call here, if just copying the file (like in the dockerfile of display) or creating this volume. I'll let you decide

GilMM27 (commented):
before approving please submit evidence this was tested

GilMM27 (commented):
I don't think it's necessary to add the extra subdir to `scripts/`, we can just place the 2 files in it but rename `entrypoint.sh` to specify that is for ollama. Also for this to work I believe it is missing adding the new yaml to `docker-compose-cpu.yaml`. How was your experience running ollama on CPU? This option wasn't available because it is not viable I believe

GilMM27 (changes requested):
Nice! Try rebasing to main so that the pr doesn't show changes that don't belong here

GerardoFJ (approved):
Lgtm on integration part :rocket:

GilMM27 (approved):
nice work!

PR #618, opened 2025-12-02
Fix integration build

biweep863:
- Add symlink
- Ignore xarm package


GilMM27 (approved):
👀

GilMM27 (changes requested):
Image

this doesn't seem to be working

PR #623, opened 2025-12-11
migrate display to ROS package

GilMM27:
ignore changes to the run.sh from other areas, that belongs to #624, I just messed up and committed it here.

Uses ros bridge to automatically forward messages through a websocket, removing the necessity of having a backend.

Thanks to RosLib, we can connect to the websocket easily and "subscribe" to topics.

Due to the low amount of dependencies needed for the display, I decided to integrate it to the ros container and run it as a ros package.

### DIFF
--- a/docker/hri/dockerfiles/Dockerfile.ROS
+++ b/docker/hri/dockerfiles/Dockerfile.ROS
@@ -1,10 +1,23 @@
 ARG BASE_IMAGE
 FROM ${BASE_IMAGE}
 
-RUN apt-get install -y python3-pyaudio
+RUN apt-get update
+RUN apt-get install -y \
+    python3-pyaudio \
+    alsa-base \
+    alsa-utils \
+    gstreamer-1.0 \
+    libasound-dev \
+    portaudio19-dev \
+    libportaudio2 \
+    libportaudiocpp0 \
+    pulseaudio \
+    curl \
+    ros-humble-web-video-server \
+    ros-humble-rosbridge-server
 
-RUN apt-get update && apt-get install -y alsa-base alsa-utils gstreamer-1.0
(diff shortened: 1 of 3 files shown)


docker/hri/dockerfiles/Dockerfile.ROS
```diff
@@ -23,4 +23,14 @@ RUN sudo usermod -aG audio ros
 
 RUN sudo chown -R ros /usr/lib/python3.10/
 RUN sudo chown -R ros /usr/local/lib/python3.10/
-RUN sudo pip install baml-py==0.86.1
\ No newline at end of file
+RUN sudo pip install baml-py==0.86.1
+
+USER root
```
Oscar-gg: Above this line, we have: `USER ros`

So now, the default user will be root when creating the container.

Are there any relevant changes introduced by this?
(I'm assuming root is needed to install these dependencies. I think we can either install them above the  `USER ros` line or add another `USER ros` below if we don't want the previous behavior to change. However, maybe there are no relevant changes and this is fine).

Oscar-gg (approved):
Nice, it seems this pr removes a lot of code duplication for display.

Stamping to unblock, but added some q above env compatibility

docker/hri/dockerfiles/Dockerfile.ROS
GilMM27: since the hri-ros.yaml specifies the user, it actually uses the ros user. Also, valid concern but install works on orin too!

Oscar-gg: > since the hri-ros.yaml specifies the user, it actually uses the ros user

I see, makes sense.

> I improved Dockerfile.ROS, I'll merge it tomorrow to give you time to check it out


Took another look, lgtm 👍

PR #624, opened 2025-12-13
fix start rerunning initialization command

GilMM27:
It was discovered that running `docker compose  start` started the container but ran in the background the `command` with which it was created. For example, if I created the container with a build command, every time the container gets started it will build in the background again. To avoid this instead we should run `up` which checks if the compose changed and recreates the container in such case, if the container is already running and it didn't change it does nothing.

Also moved the assignation of `COMMAND` below to avoid running it twice on the first if block


GerardoFJ (approved):
LGTM 🚀

PR #636, opened 2026-01-05
Add new file for tracking node using Norfair

brbrbre:
Create a new script for a tracker node using Norfair instead of ByteTrack


JocelynVelarde (commented):
Run the following commands on CMD so you don’t get the build errors when committing
`pre-commit install`
`pre-commit run —all-files`

PR #641, opened 2026-01-15
Trash detection

DanHeGa:
Trash detection node implementation using YOLO-E.


vision/packages/vision_general/scripts/trash_detection_node.py
```diff
@@ -0,0 +1,177 @@
...
+            
+            if self.current_image is not None:
+                h, w = self.current_image.shape[:2]
+                
+                for detection in self.latest_detections.detections:                    
+                    obj_detection = ObjectDetection()
+                    obj_detection.label_text = detection.label_text
+                    obj_detection.score = detection.score
+                    
+                    # Convert normalized coordinates to pixels
```
DanaeSG: I don´t think the conversion is necessary.

PR #646, opened 2026-01-20
Add baml native test and remove depricated files for nlp tests

GilMM27:


hri/packages/nlp/test/metrics/case_insensitive_exact_match.py
Oscar-gg: Are files in this package not used anymore?
What is the alternative that is being used?

Oscar-gg: nvm, I didn't see your other pr

hri/packages/nlp/test/baml_src/tests_command_interpreter.baml
```diff
@@ -1,302 +0,0 @@
-/// Tests for command interpreter functionality
```
Oscar-gg: Do we still have some baml "native" tests somewhere? I think these can still be useful to test baml specific features

GilMM27: should I just keep this file? jeje

and how would you recommend running it since I'm trying to integrate the tests to the task manager

PR #659, opened 2026-02-15
Manipulation/distance mask

Fernando94654:
- Added max_distance and min_distance to PickParams to build a distance mask

- Implemented point_in_range helper in perception_utils.py

- Updated keyboard_input.py to display only objects within the configured distance range

- Added PICK_MAX_DISTANCE constant and set it as the default value for PickParams.max_distance.

To test keyboard_input.py with the distance mask but still show all objects and see how ManipulationAction works with the mask, simply remove the "and point_in_range" in objects_callback.


GerardoFJ (changes requested):
Change parameter implementation

GerardoFJ (approved):
LGTM🚀

PR #664, opened 2026-02-17
Add word confidences to hear

GilMM27:


Oscar-gg (approved):
LGTM overall. Added a question about word confidences' computation

PR #669, opened 2026-02-21
updated ask_and_confirm

GilMM27:
Updated `find_closest` to also return the similarity rate. Updated ask and confirm to dynamically determine if asking a confirmation is necessary: 
* if there is an exact match with the provided options.
* if there is a high similarity match with one of the options.
* if the transcription confidence for the word is high.

### DIFF
--- a/task_manager/scripts/subtask_managers/hri_tasks.py
+++ b/task_manager/scripts/subtask_managers/hri_tasks.py
@@ -694,7 +694,7 @@ class HRITasks(metaclass=SubtaskMeta):
                                 target_info = closest.results[0]
                                 similarity = closest.similarities[0]
 
-                    # Skip confirmation depending on the similarity to an option if options are provided or on transcription confidence
+                    # Skip confirmation depending on the similarity to an option if options are provided and/or on transcription confidence
                     if (
                         similarity > SKIP_CONFIRMATION_SIMILARITY_THRESHOLD
                         and word_confidences.get(target_info, 0)


task_manager/scripts/subtask_managers/gpsr_single_tasks.py
```diff
@@ -138,7 +138,7 @@ def pick_object(self, command: PickObject):
                 break
 
         labels = self.subtask_manager.vision.get_labels(detections)
-        s, object_to_pick = self.subtask_manager.hri.find_closest(labels, command.object_to_pick)
+        s, object_to_pick, _ = self.subtask_manager.hri.find_closest(labels, command.object_to_pick)
```
Oscar-gg: If we'll be returning more than 2 values for many functions, perhaps we could refactor the functions to return an object instead, so that we don't need to modify all calls when adding/removing a value.

i.e. have an object like 
```python3
FindClosest:
   Status
   object_to_pick
   similarities
   ...
```

task_manager/scripts/subtask_managers/hri_tasks.py
```diff
@@ -659,31 +660,35 @@ def ask_and_confirm(
...
-                            target_info = self.find_closest(options, target_info)[0]
-                except Exception as e:
-                    print("Failed options:", e)
+                            s, closest_match, similarities = self.find_closest(options, target_info)
+                            if s == Status.EXECUTION_SUCCESS:
+                                target_info = closest_match[0]
+                                similarity = similarities[0]
+
+                    if similarity > 0.5 and word_confidences.get(target_info, 0) > 0.5:
+                        skip_confirmation = True
```
Oscar-gg: Maybe nest this below 
```python3
  if s == Status.EXECUTION_SUCCESS:
                                target_info = closest_match[0]
                                similarity = similarities[0]
```
as `similarity` depends on `s == Status.EXECUTION_SUCCESS`?

Oscar-gg: Also, maybe we can move the magic numbers (`similarity > 0.5 and word_confidences.get(target_info, 0) > 0.5`) to `hri_constants.py` or somewhere similar.

task_manager/scripts/subtask_managers/hri_tasks.py
```diff
@@ -659,31 +660,35 @@ def ask_and_confirm(
...
                     s, target_info = self.extract_data(query, interpreted_text, context)
-
                 try:
                     if options is not None:
                         foundExact = False
                         for option in options:
                             if option.lower() in target_info.lower():
                                 target_info = option
                                 foundExact = True
+                                skip_confirmation = True
```
Oscar-gg: This now implies that if a word was heard exactly as the option (even if it had low `word_confidence`) we skip the confirmation, right?

asking to double-confirm that this is desired behavior

Oscar-gg (approved):
LGTM, added minor comments/questions.

task_manager/scripts/subtask_managers/hri_tasks.py
```diff
@@ -659,31 +660,35 @@ def ask_and_confirm(
             start_time = self.node.get_clock().now()
 
             self.say(question)
-            hear_status, interpreted_text, _ = self.hear(hotwords=hotwords)
+            hear_status, interpreted_text, word_confidences = self.hear(hotwords=hotwords)
 
             if hear_status == Status.EXECUTION_SUCCESS:
                 target_info = interpreted_text
+                similarity = 1
```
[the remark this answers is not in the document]
GilMM27: if we don't provide options the dynamic skip would never happend, so in does cases the only "real" check should be word confidence. Lmk if this makes sense

Oscar-gg (approved):
LGTM

PR #673, opened 2026-02-21
Nlp test

luisrosales852:
It passes all the tests

### DIFF
--- a/hri/packages/nlp/scripts/extract_data.py
+++ b/hri/packages/nlp/scripts/extract_data.py
@@ -124,7 +124,7 @@ class DataExtractor(Node):
                 )
                 if response.result == "":
                     self.get_logger().error(
-                        f"No location found in {request.full_text}. Returning empty string as result."
+                        f"No location found in {request.full_text}. Returning empty string as a result."
                     )
             return response
 


hri/packages/nlp/test/data_extractor.json
GilMM27: last I remember there is only support for tags `drink` and `name`. For everything else the `LLM_` prefix is needed

hri/packages/nlp/scripts/extract_data.py
GilMM27: I think it's cleaner if you fallback here. if the function didn't work then don't return and continue bellow which uses the llm

hri/packages/nlp/test/data_extractor.json
luisrosales852: There's only support for loc and name, for everything else LLM prefix is needed

hri/packages/nlp/scripts/llm_utils.py
```diff
@@ -390,7 +390,13 @@ def get_most_likely_label(self, text):
 
         # Get the index of the maximum score
         max_index = scores.index(max(scores))
-        return labels[max_index]
+        max_label = labels[max_index]
+        idk_index = labels.index("i don't know")
+        if (max_label == "yes" or max_label == "no") and (
+            (scores[max_index] - scores[idk_index]) < 0.01
+        ):
+            return ""
```
GilMM27: I see now, you're right lol

I think it's okay to return i don't know, just make sure that response is handled correctly everywhere

sorry for the confusion

GilMM27 (changes requested):
almost done hehe

hri/packages/nlp/scripts/extract_data.py
```diff
@@ -115,6 +115,19 @@ def extract_info_requested(
...
+                    "No location found in text using spacy. Attempting to extract location using LLM."
+                )
+                response.result = self.extract_via_llm(
+                    request.full_text, "location", ""
+                )
+                if response.result == "":
+                    self.get_logger().error(
+                        f"No location found in {request.full_text}. Defaulting to returning same text"
+                    )
+                    response.result = ""
```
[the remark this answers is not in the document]
GilMM27: also pls fix the error log

PR #681, opened 2026-02-25
Vision/customer detect

Fernando94654:
Added sitting recognition to detect customers, in addition to hand-raise detection. The system first uses YOLO, and if it fails, it falls back to MoonDream.
Screenshot from 2026-02-23 20-46-15
Screenshot from 2026-02-23 20-45-46
Screenshot from 2026-02-23 20-45-23
Screenshot from 2026-02-23 20-44-50
Screenshot from 2026-02-23 17-05-17
Screenshot from 2026-02-23 16-53-33
Screenshot from 2026-02-23 16-53-06
Screenshot from 2026-02-23 16-52-06
Screenshot from 2026-02-23 16-51-10
Screenshot from 2026-02-23 16-50-15
Screenshot from 2026-02-23 16-49-33
Screenshot from 2026-02-23 16-46-37


vision/packages/moondream_run/scripts/moondream_node.py
```diff
@@ -195,6 +200,48 @@ def crop_query_callback(self, request, response):
 
         return response
 
+    def is_sitting_callback(self, request, response):
```
DanaeSG: Instead of creating a new service and function in Moondream, you can directly call the crop query within the customer node. Simply pass the cropped frame as a parameter and handle the “yes” or “no” response logic inside that same script.

DanaeSG (changes requested):

vision/packages/vision_general/scripts/customer_node.py
```diff
@@ -230,21 +234,64 @@ def get_customer_callback(self, req, res):
...
+            self.get_logger().warn("Service call timed out")
+            return False
+
+        if future.exception() is not None:
+            self.get_logger().error(f"Service call failed: {future.exception()}")
+            return False
+
+        result = future.result()
+        if result is None or not result.success:
             return False
```
DanaeSG: Change it for 
```
        future = self.moondream_client.call_async(request)
        future = self.wait_for_future(future, 15)
        result = future.result()
        
        if result is None:
        if result.success:
```

Where wait_for_future:
```
    def wait_for_future(self, future, timeout=5):
        start_time = time.time()
        while future is None and (time.time() - start_time) < timeout:
            pass
        if future is None:
            return False
        while not future.done() and (time.time() - start_time) < timeout:
            pass

        return future
```

`rclpy.spin_until_future_complete` doesn´t work in this context because you are calling another node and it would enter an infinite loop if used (as it is expecting the main node to complete as well).

DanaeSG (changes requested):

DanaeSG (approved):

PR #724, opened 2026-03-07
Manipulation/hand2hand

Fernando94654:
- Add GoToHand action for the HRI Challenge to position the gripper close to the person's hand (within 10 cm).

- Tested using an RViz debug point placed at the person's hand.
Video demostration:

https://github.com/user-attachments/assets/4795ed10-16ae-4f0a-8f66-257952a7f439


JLDominguezM (approved):
Good PR 👍

PR #745, opened 2026-03-12
Hri/change tts image

Fernando94654:
This PR updates the TTS container base image.

**Main changes**
- Switched the TTS base image to `python:3.10-bookworm`, avoiding the use of the RoboRegros base image that included ROS 2 unnecessarily.
- Added explicit Hugging Face/Transformers cache environment variables in `tts.yaml`:
  - `HOME=/app`
  - `HF_HOME=/app/.cache/huggingface`
  - `HUGGINGFACE_HUB_CACHE=/app/.cache/huggingface/hub`
  - `TRANSFORMERS_CACHE=/app/.cache/huggingface/transformers`
  - `XDG_CACHE_HOME=/app/.cache`

**Why**
The container runs as a non-root user and the default cache path was resolving to `/.cache`, causing:
`PermissionError: [Errno 13] Permission denied: '/.cache'`.

**Expected impact**
- Prevents Hugging Face cache writes to non-writable system paths.
- Avoids using the RoboRegros ROS 2 image when it is not required.
- Makes Kokoro TTS startup more reliable across environments (including L4T).

No API or behavior changes to the TTS service.

### DIFF
--- a/docker/hri/compose/tts.yaml
+++ b/docker/hri/compose/tts.yaml
@@ -6,7 +6,7 @@ services:
       context: ../../..
       dockerfile: docker/hri/dockerfiles/Dockerfile.tts
       args:
-        BASE_IMAGE: roborregos/home2:${ENV_TYPE:-cpu}_base
+        BASE_IMAGE: ${TTS_BASE_IMAGE:-python:3.9-bookworm}
     volumes:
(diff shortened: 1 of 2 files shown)


Oscar-gg (approved):
LGTM

PR #756, opened 2026-03-17
714 move microservices from scripts

Didi-121:
## Changes:

- Microservices Relocation: The STT and TTS microservices were relocated to the hri/microservices/ directory to cleanly separate the machine learning dependencies from the ROS2 orchestration layer.

- Standalone Proto Package: A standalone Python package named proto_interfaces was created to manage the import paths for the generated Protocol Buffer modules. This approach leverages standard Python packaging to replace runtime directory manipulation (such as sys.path.append). It is an elegant solution because it establishes a single, installable source for the communication schemas. As a result, both the ROS2 nodes and the independent microservices can utilize standard, explicit imports (e.g., from proto_interfaces import speech_pb2), which ensures reliable path resolution, enables IDE autocomplete, and maintains a strict architectural boundary.

### Cuda build: 
Screenshot from 2026-03-17 13-42-44

### Cuda tts:
Screenshot from 2026-03-17 13-41-07
 
Screenshot from 2026-03-17 13-42-00

### Cuda  stt: 
Screenshot from 2026-03-17 14-04-01

### Orin build 
image

### Orin tts
image


docker/hri/dockerfiles/Dockerfile.stt
```diff
@@ -2,7 +2,10 @@ FROM python:3.10-bookworm
 
 WORKDIR /app
 
-COPY ../../hri/requirements/stt.txt requirements.txt
+COPY hri/proto_interfaces /tmp/proto_interfaces
+RUN pip install --no-cache-dir /tmp/proto_interfaces
```
Oscar-gg: Is this necessary?

hri/packages/speech/scripts/hear_streaming.py
```diff
@@ -17,10 +15,7 @@
 
 from frida_interfaces.action import SpeechStream
 from frida_interfaces.msg import AudioData
-
-sys.path.append(os.path.join(os.path.dirname(__file__), "stt"))
-import speech_pb2
-import speech_pb2_grpc
+from proto_interfaces import speech_pb2, speech_pb2_grpc
```
Oscar-gg: Nice

Oscar-gg (commented):
overall LGTM, added some questions that I think need to be addressed before merge.

hri/proto_interfaces/proto_interfaces/tts_pb2.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
```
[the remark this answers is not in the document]
Didi-121: The files marked as new are the exact same files but with one difference in the import (from . import tts_pb2 as tts__pb2). The 4 proto files in microservices are leftovers from the refactor

Oscar-gg (approved):
Nice, LGTM

PR #768, opened 2026-03-17
made using llm the default

luisrosales852:
done a while ago but yeah forgot to do the pull request. Basically building up to what I did in the llm fallback part. It guarantees that any item that is not loc or name will go to an llm fallback immediately, while loc or name will go to their respective methods and if they fail then llm fallback regardless. If you want to add any new methods its just an if statement before the general llm fallback


Oscar-gg (approved):
LGTM, approving, but apply copilot's comment

PR #770, opened 2026-03-18
Support hot

luisrosales852:
Haven’t fully tested yet but leave it here. Don’t approve


Oscar-gg (commented):
Overall looks good. Waiting for testing before approving

PR #775, opened 2026-03-19
Refactor run scripts to utilize common flag parsing

JocelynVelarde:
## extract shared flags and env setup from area run.sh scripts

Moved duplicated boilerplate from all 5 area `run.sh` scripts into two new functions in `lib.sh`:

- `parse_common_flags` : handles `-d`, `--build`, `--build-image`, `--upload-image`, `--recreate`, `--down`, `--stop`
- `setup_common_env "<area>"` : resets `.env`, writes `CYCLONE_INTERFACE`, `LOCAL_USER_ID`, `LOCAL_GROUP_ID`, `BASE_IMAGE`, `IMAGE_NAME`, and creates `install/`, `build/`, `log/`

### No behavior changes

All flags work exactly the same as before. This is a pure refactor, no new functionality, no changes to Docker commands, compose files, or environment variables.

### Notes for reviewers

- `navigation` now resolves `COMPOSE_FILE` based on `ENV_TYPE` **before** parsing flags, so `--down/--stop/--recreate` correctly target the right compose file on gpu/l4t environments. This was a latent bug in the original.
- `hri` uses `compose/.env` instead of `.env`, so it has a local inline equivalent of `setup_common_env` writing to that path. The rest of its logic (display setup, model download, db regeneration) is unchanged.
- `manipulation` now has `BASE_IMAGE` and `IMAGE_NAME` written to `.env` via `setup_common_env`, which the original was missing. This aligns it with the other areas and is needed for `--build-image` to work correctly.


GilMM27 (changes requested):

GilMM27 (approved):
soft approve

JLDominguezM (changes requested):
Nope

JLDominguezM (approved):
LGTM 🤔

PR #776, opened 2026-03-19
Dynamic Parameters Nav for Lidar and Stm32

Bandarcrono:
- Added safety measure so that if the Lidar or Stm32 does not get recognized/connected the container cannot be run-
- Added a dynamic way to assign the Lidar and Stm32 port based on the serial number (which is unique for each device)
- Created a [[setup-USB.sh](http://setup-usb.sh/)](http://setup-usb.sh/) script to assign symlinks names abbreviations, so that we are not relying on a specific port name.
- Gave permissions to the docker compose to search and make modifications to the ports and made self check if the permissions are already made.
- Added self check status for the Lidar and Dashgo and if it is not sending data correctly it is going to eliminate the launch of nav_basics.

# Results 
image
image


GerardoFJ (dismissed):
Good pr just some necessary changes

CamilaTite26 (approved):

PR #801, opened 2026-03-26
add bag pose

Fernando94654:


GilMM27:
I think this was added in #810

GilMM27:
amend or close PR @Fernando94654

PR #805, opened 2026-03-26
feat: Add frame counting and skipping functionality in zero_shot_obje…

JLDominguezM:
Fix bug in zero shot detections node


CamilaTite26 (approved):
lgtm

PR #814, opened 2026-03-27
fix speech pipeline (initial prompt + hotwords)

luisrosales852:


Oscar-gg (commented):
LGTM, can you apply copilot's comments that make sense, and add a screenshot of the test? (in the test compare the difference between using the initial_prompt and not using it please if possible)

GilMM27 (changes requested):
looking good. Please test that the pipeline is correctly receiving the values and transcription is still working correctly for an approve

PR #817, opened 2026-03-28
Update Shm on dockers ( Added Nav)

GerardoFJ:
Change Log:

  - Increase IOX_INTERNAL_MAX_NUMBER_OF_NOTIFIERS from 256 to 1024 in all 8 Dockerfiles     
  (roudi, navigation x3, zed, vision x3). This was hardcoded in iceoryx v2.0.6 and limited
  how many subscribers can attach to a single listener. Both nav2_container and             
  rtabmap_container race for the same pool  whichever exhausts it first fails, causing the
  intermittent Failed to attach iox subscriber to iox listener errors and random
  bt_navigator / rtabmap startup failures.
  - Increase IOX_MAX_CHUNKS_HELD_PER_SUBSCRIBER_SIMULTANEOUSLY to 2048 (already in
  Dockerfiles from prior work). Fixes TOO_MANY_CHUNKS_HELD_IN_PARALLEL flooding during      
  startup when ~20+ nodes all publish rt/parameter_events simultaneously.
  - Disable autostart on map_service in navigation_composition.launch.py. Previously        
  map_service self-transitioned to active before the lifecycle manager could manage it,     
  causing No transition matching X found for current state active errors on every launch.


docker/navigation/Dockerfile.cpu
```diff
@@ -14,10 +14,45 @@ RUN cd /workspace/src && sudo apt update && sudo rosdep init && rosdep update
...
+    git clone --depth 1 -b releases/0.10.x https://github.com/eclipse-cyclonedds/cyclonedds.git && \
+    cd cyclonedds && mkdir build && cd build && \
+    . /opt/ros/humble/setup.sh && \
+    cmake .. \
+        -DCMAKE_INSTALL_PREFIX=/opt/ros/humble \
+        -DCMAKE_PREFIX_PATH=/opt/ros/humble \
+        -DENABLE_SHM=ON \
+        -DCMAKE_BUILD_TYPE=Release && \
+    make -j$(nproc) && sudo make install && \
+    cd / && rm -rf /tmp/cyclonedds
```
JLDominguezM: Lets try to not repeat this in all dockerfiles, extract this into an base imagen in other PR

JLDominguezM (approved):

PR #818, opened 2026-03-28
add image support for tts in cuda and l4t

Fernando94654:
Everything is still working fine.
.Screenshot from 2026-03-28 15-07-08
Screenshot from 2026-03-28 15-05-52


docker/hri/dockerfiles/Dockerfile.tts-l4t
```diff
@@ -0,0 +1,43 @@
+FROM dustynv/l4t-pytorch:r36.4.0
+
```
Oscar-gg: Do we need this file? Can't we use the same as the other tts file and only change the base image

Oscar-gg (approved):
Lgtm

PR #832, opened 2026-04-02
Manip improve pick

JLDominguezM:
### GPD Speed (77x faster grasp generation)
- `num_orientations` 24/32 → 8, `num_finger_placements` 64/32 → 10 in both .cfg
- GPD runtime: ~5-8s → ~0.09-0.15s per config

### Point Cloud Quality
- `remove_outliers` = 0 → 1 in both .cfg (enables PCL StatisticalOutlierRemoval)
- `camera_position` = 0 0 0 → 0 0.06 1.17 in both .cfg (correct normal orientation)
- `simplificator_factor` = 3 → 1 in down_sample_pc.cpp (3x denser point cloud)
- `max_update_rate` = 0.5 → 2.0 in sensors_3d.yaml (4x faster octomap refresh)

### Feasibility Check Fix
- `PICK_MIN_HEIGHT` = 0.1 → 0.04 in manipulation_constants.py (allows grasps on objects <10cm)
- `thresh_rad` = 0.2/0.15 → 0.52 in both .cfg (accepts 30° approach deviation vs 11°)

### Code Cleanup
- Removed duplicate `wait_for_service()` in grasp_utils.py
- Gripper close loop `range(2)` → single call in PickManager.py
- Sleep between GPD configs 1.0s → 0.2s in PickManager.py
- Busy-wait `pass` → `time.sleep(0.001)` in ros_utils.py (frees CPU core)


Fernando94654 (approved):
LGTM!

PR #845, opened 2026-04-02
Hri - change debug path

biweep863:
- add debug to config
- change save path


hri/packages/speech/config/microphone.yaml
```diff
@@ -4,4 +4,4 @@ audio_capturer:
     MIC_INPUT_CHANNELS: 32
     MIC_OUT_CHANNELS: 32
     RAW_AUDIO_TOPIC: REPLACE
-    DEBUG: false
+    DEBUG: False
```
Oscar-gg: Why uppercase? Does it work only in uppercase?

Oscar-gg (approved):

PR #848, opened 2026-04-03
feat: add colored logger for task managers with subsystem-based ANSI …

JLDominguezM:
## Here we have some examples
image
image


GerardoFJ (approved):
LGTM :rocket:

PR #862, opened 2026-04-07
Hri tune stt

GilMM27:
* fixed skip confirm for capitalized and punctuated words
* adjusted thresholds for harsher voice detection


task_manager/task_manager/subtask_managers/hri_tasks.py
```diff
@@ -410,7 +410,7 @@ def _get_keyword(self, msg: String) -> None:
     def hear(
         self,
         hotwords="",
-        silence_time=4.0,
```
Oscar-gg, line 413: I'd suggest to propagate it to ask_and_confirm

Oscar-gg (approved):

task_manager/task_manager/subtask_managers/hri_tasks.py
GilMM27: wdym?

PR #870, opened 2026-04-09
Vision/follow face rot

Fernando94654:
For certain xArm positions, the camera is 180° rotated, so this PR adds a topic to rotate the image and run face recognition appropriately. Tested with
ros2 topic pub --once /vision/face_recognition/flip_image std_msgs/msg/Bool "{data: false}"


DanaeSG (approved):
lgtm

PR #871, opened 2026-04-09
hooked up initial prompt to speech backend

luisrosales852:
Final speech pipeline fix, noticed small changes


GilMM27 (approved):
LGTM, please also commit the generated python scripts from proto

PR #877, opened 2026-04-10
Vision/detect trash

DanHeGa:
- Add trash detections with category-based filtering
- Add trashcan point detection


DanaeSG (approved):
lgtm!

PR #884, opened 2026-04-12
Hri - new retrained oww model

biweep863:
- edge impulse model


GilMM27 (changes requested):
too many false positives :(

GilMM27 (changes requested):
remove model.eim models pls

GilMM27 (approved):
PRIME

PR #894, opened 2026-04-13
added node for inferencing edge impulse

GilMM27:
this is for the knock / doorbell model. Model needs some tuning still


biweep863 (approved):
lgtm

PR #910, opened 2026-04-15
Urdf collision update

AleGonzcamilla:
Change the collision of the URDF into a simplified mesh that included the new size of  the speaker and screen.


GilMM27 (approved):
approving to unblock

PR #912, opened 2026-04-17
hric tmr 2026

GilMM27:
* Skip bag stuff in hric
* split display from hri so it's easier to run in different orins so that we take advantage of roudi


Oscar-gg (approved):
LGTM

PR #922, opened 2026-04-23
Sim: load house mesh with CoACD collisions + macro-based object pool

JLDominguezM:
## Summary

- Re-enable `house.xacro` in the MuJoCo sim launch so the TMR2025 scene (walls, furniture, tables) is visible again.
- Pre-decompose `house.STL` with CoACD (62 convex pieces, threshold=0.05) so MuJoCo collision geometry matches what is visually there: doorways and gaps are passable, walls and furniture still block.  `xacro2mjcf`'s existing `add_composite_collisions` auto-wires the pieces; `house.xacro` itself stays unchanged.
- Replace the static `table_and_object.xacro` fixture (hardcoded table + cube) with `object_pool.xacro`: a macro-based pool of freejoint objects parked behind the robot.  Adding an object is one line; moving it at runtime is `Ctrl+Shift+click-drag` in the MuJoCo viewer.

## Test plan
- [x] `colcon build --symlink-install --packages-select frida_description mujoco_spawn`
- [x] Relaunch `mujoco_sim_init.launch.py`; confirm the house renders and the three parked cubes (red / green / blue) are visible behind the robot.
- [x] Drag a cube through a doorway in the MuJoCo viewer — should pass through the opening instead of being stopped by an invisible wall.
- [x] Drop a cube on top of a table — should rest, not tunnel.
- [x] Verify realtime factor is still >= 0.9 (62 extra convex colliders is the main perf concern).


biweep863 (approved):
good

PR #923, opened 2026-04-23
Sim: stop launching vision nodes from the manipulation container

JLDominguezM:
## Summary
- \`mujoco_sim_init.launch.py\` was starting \`object_detector_2d\`'s zero-shot detector and an inline \`/vision/detections\` relay, but \`object_detector_2d\` only exists in the vision container.  On a fresh manipulation container the launch dies with \`package 'object_detector_2d' not found\`.
- This PR removes both entries from the sim launch so it matches the real-robot workflow: the manipulation container brings up the arm, MoveIt, physics and the self-filtered pointcloud only; vision is launched separately in the vision container (unchanged).

## Test plan
- [x] Fresh manipulation container (without vision_general or object_detector_2d built) launches cleanly: \`ros2 launch mujoco_spawn mujoco_sim_init.launch.py\` — no "package not found".
- [x] MuJoCo physics + MoveIt + RViz + self-filtered cloud come up as before.
- [x] Start vision separately in the vision container: \`ros2 launch object_detector_2d zero_shot_object_detector_node.launch.py use_sim_time:=true\` — /vision/zero_shot_detections publishes and pick_and_place in terminal 2 still works once a \`zero_shot -> /vision/detections\` relay is running somewhere (follow-up: document in a README or add a small vision-side launch arg).

### DIFF
--- a/manipulation/packages/mujoco_spawn/launch/mujoco_sim_init.launch.py
+++ b/manipulation/packages/mujoco_spawn/launch/mujoco_sim_init.launch.py
@@ -11,7 +11,6 @@ from launch_ros.actions import Node
 from launch_ros.substitutions import FindPackageShare
 from launch.event_handlers import OnProcessStart
 from launch.actions import (
-    ExecuteProcess,
     IncludeLaunchDescription,
     OpaqueFunction,
     RegisterEventHandler,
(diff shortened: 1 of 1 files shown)


GerardoFJ (approved):
LGTM 🚀

PR #926, opened 2026-04-24
feat: tune hotwords

CamilaTite26:


task_manager/scripts/hric_task_manager.py
```diff
@@ -63,6 +67,30 @@ def __init__(self):
         self.running_task = True
         self.message = ""
 
+        self.drink_options = ["water", "coke", "coffee", "tea"]
+        try:
+            objects_path = os.path.join(
```
biweep863: use frida constants

task_manager/scripts/hric_task_manager.py
```diff
@@ -63,6 +67,30 @@ def __init__(self):
...
+                get_package_share_directory("frida_constants"), "data", "objects.json"
+            )
+            with open(objects_path, "r") as f:
+                data = json.load(f)
+                if "drink" in data.get("categories", {}):
+                    self.drink_options = data["categories"]["drink"]
+        except Exception:
+            pass
+
+        self.name_options = [
```
biweep863: i think it will be accurate to use the data from frida_contants

task_manager/task_manager/subtask_managers/hri_tasks.py
```diff
@@ -658,6 +656,19 @@ def ask_and_confirm(
             Status: the status of the execution
             str: answer to the question
         """
+        # Auto-configure STT context using options
+        if options is not None:
```
biweep863, line 659: why options is used inthis case?

GilMM27:
I think you should merge main and improve the current (merged) implementation for this. Also remember to test how well does it work!

PR #941, opened 2026-05-11
Deprecate and fix categorization

GilMM27:
* Deprecated dynamic categorization via LLM
* `categorize_objects` refactored to:
1. Analyzes Current Shelves: It first scans all shelves to see what they already contain and
     maps existing objects to their semantic categories (e.g., "Shelf 1 has 'apple', so it's
     for 'fruit'").
2.  Groups Table Objects: It takes the list of objects from the table and groups them by
     their specific categories.
3. Prioritizes Existing Categories: If a table object belongs to a category already present
     on a shelf, it is automatically assigned to that same shelf.
4. Distributes New Categories:
      4.1. To Empty Shelves: If a category is new, it looks for empty shelves and distributes
         them using a "round-robin" approach to keep things organized.
       4.2. To Occupied Shelves (Overflow): If no empty shelves are left, it distributes the
         remaining new categories across all available shelves.
5. Avoids Generic Labels: Instead of using "miscellaneous," it returns the actual specific
     category names.
6. Multi-Category Support: It allows a single shelf to hold multiple categories (e.g.,
     "fruit food"), providing the robot with a precise list of what is stored where.
* try catch block for extract data requests, before the node just crashed


Oscar-gg:
Depricated -> Deprecated

hri/packages/nlp/test/categorize_objects.json
```diff
@@ -35,7 +35,7 @@
         "answer": {
             "0": {"category": "drink", "objects_to_add": ["milk", "fanta"]},
             "1": {"category": "dish", "objects_to_add": ["fork"]},
-            "2": {"category": "miscellaneous", "objects_to_add": ["tangerine", "ketchup"]}
+            "2": {"category": ["food", "fruit"], "objects_to_add": ["tangerine", "ketchup"]}
```
Oscar-gg, line 38: Maybe we could always have an array (even for shelves with one category)?

Oscar-gg (approved):
Nice, LGTM 🚀

PR #971, opened 2026-05-14
docs for AI agents

GilMM27:
added ./scripts dir and moved other sh from root to there. Updated docs accordingly
created `setup_agent.sh` which prompts the user to setup an agent from a choice of options or custom
cleaned up depricated docs


biweep863 (commented):
could you do this for multiple agents? nt just gemini

setup_cyclonedds.sh
```diff
@@ -1,212 +0,0 @@
-#!/bin/bash
-# Setup or revert CycloneDDS for ROS 2
```
Oscar-gg, line 2: Is this related to "docs for AI agents"?

status.sh
Oscar-gg: Why have this script and `setup_cyclonedds.sh` as wrappers instead of just leaving them at `scripts`?

docs/ai/architecture.md
```diff
@@ -0,0 +1,33 @@
+# FRIDA Architecture
```
Oscar-gg, line 1: I think the info here is heavily duplicated with the [docs](https://athome.roborregos.com/).

We should have a way of syncing this info from the docs (and ensuring the docs are up-to-date) or the other way around (have a way for the docs to auto update from this).

docs/ai/tech_stack.md
```diff
@@ -0,0 +1,23 @@
+# Tech Stack
```
Oscar-gg, line 1: Ditto

docs/ai/coding_standards.md
```diff
@@ -0,0 +1,51 @@
+# Coding Standards & Conventions
```
Oscar-gg, line 1: Maybe the sync wouldn't apply for very code-specific docs like this one.

scripts/setup_agent.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
```
Oscar-gg, line 1: Unsure if this is the way this is usually done. I recall seeing [this](https://github.com/facebookincubator/meta-wearables-webapp/blob/main/install-skills.sh) for the meta glasses' sdk. Leaving it here in case you want to check it out.

Oscar-gg (commented):
General comments

status.sh
GilMM27: yep mb

setup_cyclonedds.sh
GilMM27: no, but decided it would be a quick fix

docs/ai/architecture.md
GilMM27: I think docs and ai are meant for different purposes, one for humans and the other for agents. Agents will add this context to the session in exchange for tokens so we have to find the right balance of useful info without bloating. This is why I think we should keep them separate.

scripts/setup_agent.sh
GilMM27: It's kinda the same right? except they have agent specific content. Is this what you rather have?

Oscar-gg (approved):
LGTM

PR #972, opened 2026-05-15
fix(manip): restore real-robot pick after Mujoco PR regression

GerardoFJ:
## Summary
The Mujoco PR (#739) forced `FRIDA_Real.urdf.xacro` + `mujoco_plugin=true` for the real robot in `moveit_configs_builder_sim.py`. That URDF unconditionally adds a `world` link sitting 0.3 m above `base_link` via `world_to_base`, switching MoveIt's planning frame from `base_link` to `world`. On the real robot:

- The plane collision object (added in `base_link`) got stored by MoveIt in `world` with the 0.3 m offset baked in. `pick_server.check_feasibility` compared grasp poses (still in `base_link`) against the offset plane z and rejected every valid grasp as 'below acceptable height'.
- The launch default for `pick_min_height` jumped from `PICK_MIN_HEIGHT` (0.04) to 0.1, tightening the already-broken check.

This restores the Apr 16 working state for the real robot:

- `moveit_configs_builder_sim.py`: revert URDF/SRDF selection and drop the forced `mujoco_plugin=true`. Real robot is back on `FRIDA.urdf.xacro` and the conditional `.sim` SRDF, which keeps `planning_frame = base_link`. **Note:** this reverts MuJoCo sim support; sim needs a dedicated config path before it can be re-enabled.
- `pick_and_place.launch.py`: `pick_min_height` default back to 0.04.
- `pick_server.check_feasibility`: transform the candidate pose into the plane's frame via tf2 before comparing z. No-op when frames match (current real-robot case), but defensive against a future split between planning frame and GPD target_frame.

## Test plan
- [x] Apple pick on real xArm6 plans and executes after this change
- [ ] Verify no regression on cutlery picks (flat objects)
- [ ] Restore MuJoCo sim path in a follow-up PR with a sim-only builder

### DIFF
--- a/manipulation/packages/arm_pkg/arm_pkg/moveit_configs_builder_sim.py
+++ b/manipulation/packages/arm_pkg/arm_pkg/moveit_configs_builder_sim.py
@@ -285,7 +285,6 @@ class MoveItConfigsBuilder(ParameterBuilder):
             'report_type': report_type,
             'baud_checkset': baud_checkset,
             'default_gripper_baud': default_gripper_baud,
-            'mujoco_plugin': "true",
             'dof': dof,
             'robot_type': robot_type,
             'prefix': prefix,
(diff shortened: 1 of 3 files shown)


GerardoFJ:
Closing: PR creator must be JLDominguezM. Reopening from the same branch.

PR #973, opened 2026-05-15
fix(manip): restore real-robot pick after Mujoco PR regression

JLDominguezM:
The Mujoco PR (#739) forced FRIDA_Real.urdf.xacro + mujoco_plugin=true for the real robot in moveit_configs_builder_sim.py. That URDF unconditionally adds a 'world' link sitting 0.3 m below base_link via the world_to_base joint, which made MoveIt's planning frame 'world' instead of 'base_link'.

Consequences on the real robot:
- The plane collision object added in base_link was stored by MoveIt in 'world', 0.3 m above its real z. pick_server.check_feasibility compared grasp poses (still in base_link) against the offset plane z, so every valid grasp got rejected as 'below acceptable height'.
- pick_min_height launch default jumped from PICK_MIN_HEIGHT (0.04) to 0.1, tightening the same already-broken check.

Restore the Apr 16 working state for the real robot:
- moveit_configs_builder_sim.py: revert URDF/SRDF selection and drop the forced mujoco_plugin=true. Real robot is back on FRIDA.urdf.xacro and the conditional .sim SRDF, which keeps planning_frame = base_link. NOTE: this reverts MuJoCo sim support; sim needs a dedicated config path before it can be re-enabled.
- pick_and_place.launch.py: pick_min_height default back to 0.04.
- pick_server.check_feasibility: transform the candidate pose into the plane's frame via tf2 before comparing z. No-op when frames already match (current real-robot case), but defensive against a future split between planning frame and gpd target_frame.

Verified: apple pick on the real xArm6 now plans + executes after this change.

### DIFF
--- a/manipulation/packages/arm_pkg/arm_pkg/moveit_configs_builder_sim.py
+++ b/manipulation/packages/arm_pkg/arm_pkg/moveit_configs_builder_sim.py
@@ -285,7 +285,6 @@ class MoveItConfigsBuilder(ParameterBuilder):
             'report_type': report_type,
             'baud_checkset': baud_checkset,
             'default_gripper_baud': default_gripper_baud,
-            'mujoco_plugin': "true",
             'dof': dof,
             'robot_type': robot_type,
             'prefix': prefix,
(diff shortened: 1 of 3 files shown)


GerardoFJ (approved):
LGTM 🚀

PR #974, opened 2026-05-15
rm pick plane

Fernando94654:
Removed plane collision; the octomap is enough. Also reduced padding_offset to maintain octomap voxels around the objects.

https://github.com/user-attachments/assets/d9e4686d-2cce-40c7-8a13-15d91b131a7f

### DIFF
--- a/manipulation/packages/arm_pkg/config/sensors_3d.yaml
+++ b/manipulation/packages/arm_pkg/config/sensors_3d.yaml
@@ -6,7 +6,8 @@ point_cloud_camera:
     point_cloud_topic: /point_cloud
     max_range: 1.5
     point_subsample: 1
-    padding_offset: 0.02
-    padding_scale: 1.05
+    # Octomap padding: 0.03 provides the best balance for covering robot collisions
+    # without causing false collisions with the environment.
+    padding_offset: 0.03
     max_update_rate: 2.0
     filtered_cloud_topic: filtered_cloud
--- a/manipulation/packages/arm_pkg/config/sensors_3d.yaml
+++ b/manipulation/packages/arm_pkg/config/sensors_3d.yaml
@@ -6,7 +6,7 @@ point_cloud_camera:
     point_cloud_topic: /point_cloud
     max_range: 1.5
     point_subsample: 1
-    padding_offset: 0.1
+    padding_offset: 0.02
     padding_scale: 1.05
     max_update_rate: 2.0
     filtered_cloud_topic: filtered_cloud
--- a/manipulation/packages/arm_pkg/config/sensors_3d.yaml
+++ b/manipulation/packages/arm_pkg/config/sensors_3d.yaml
@@ -9,5 +9,6 @@ point_cloud_camera:
     # Octomap padding: 0.03 provides the best balance for covering robot collisions
     # without causing false collisions with the environment.
     padding_offset: 0.03
+    padding_scale: 1.05
     max_update_rate: 2.0
     filtered_cloud_topic: filtered_cloud
--- a/manipulation/packages/perception_3d/src/masive_testin.cpp
+++ b/manipulation/packages/perception_3d/src/masive_testin.cpp
@@ -216,65 +216,6 @@ public:
 
     *cloud = result->cluster;
 
-    auto req2table =
-        std::make_shared<frida_interfaces::srv::RemovePlane::Request>();
-
-    req2table->extract_or_remove = false;
-    req2table->close_point.header.frame_id = point.header.frame_id;
-    req2table->close_point.header.stamp = point.header.stamp;
-    req2table->close_point.point.x = point.point.x;
-    req2table->close_point.point.y = point.point.y;
-    req2table->close_point.point.z = point.point.z;
-
-    RCLCPP_INFO(this->get_logger(), "Sending request to remove plane");
-
-    auto res2table =
-        this->call_services_node->remove_plane_client->async_send_request(
-            req2table);
-
-    RCLCPP_INFO(this->get_logger(), "Waiting for response");
-
-    auto status2 =
-        wait_for_future_with_timeout<frida_interfaces::srv::RemovePlane>(
-            res2table, this->call_services_node->get_node_base_interface(),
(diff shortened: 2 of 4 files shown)


manipulation/packages/arm_pkg/config/sensors_3d.yaml
```diff
@@ -6,7 +6,7 @@ point_cloud_camera:
     point_cloud_topic: /point_cloud
     max_range: 1.5
     point_subsample: 1
-    padding_offset: 0.1
+    padding_offset: 0.02
```
JLDominguezM: Could you add a short comment justifying this value? Going from 0.1 to 0.02 is a 5× reduction and without context someone is going  to "fix" it back later.

manipulation/packages/perception_3d/src/masive_testin.cpp
JLDominguezM, line 57: The RemovePlane cleanup is half-done here. The calls were removed from the pick flow, but:
 - The remove_plane_client member declaration and create_client / wait_for_service(30s) are still in place.

JLDominguezM, line 68: .

JLDominguezM, line 79: .

JLDominguezM, line 288: There is still a live RemovePlane call inside place_service

JLDominguezM: If the goal of this PR is to deprecate the RemovePlane service, please finish removing it here too. If not, that is fine but note that the constructor will block the node for 30 s on startup whenever the RemovePlane service isn't running.

manipulation/packages/pick_and_place/pick_and_place/pick_server.py
JLDominguezM, line 1: Heads-up: check_feasibility was acting as a cheap belt-and-suspenders against grasps falling below the table. With it gone, the entire safety net is MoveIt + the octomap. If for any reason the octomap doesn't contain the table at the moment of planning (occlusion, late TF, dropped frames, sensor restart), we no longer have a fallback that rejects the grasp.

JLDominguezM: Would it be worth keeping a minimal floor-guard based on base_link.z (or a fixed world-z threshold), independent of the plane primitive? It would cost almost nothing and cover the failure mode where the octomap is momentarily empty.

manipulation/packages/pick_and_place/pick_and_place/pick_server.py
```diff
@@ -711,13 +683,6 @@ def wait_for_future(self, future):
...
         self.collision_objects = self.get_collision_objects()
 
-    def find_plane(self):
-        self.plane = None
-        for obj in self.collision_objects:
-            if PLANE_NAMESPACE in obj.id:
-                self.get_logger().info(f"Plane object found: {obj.id}")
-                self.plane = obj
-
     def attach_pick_object(self):
```
JLDominguezM, line 686: Previously this method dropped any collision object that sat below the detected plane (if self.plane is not None and  self.object_in_plane(obj, self.plane)). With the plane gone, ghost objects from perception that show up below the table will now stay in the planning scene and could end up attached to the EEF.
Can you confirm that the perception side no longer produces those false positives? If it still does, we'll need an equivalent filter (e.g., absolute z threshold) before this is safe to merge.

manipulation/packages/pick_and_place/launch/pick_and_place.launch.py
```diff
@@ -42,11 +42,6 @@ def generate_launch_description():
                 emulate_tty=True,
                 parameters=[sim_time_param],
             ),
-            DeclareLaunchArgument(
-                "pick_min_height",
-                default_value="0.04",
-                description="Minimum height (m) a grasp pose must clear above the detected support plane. Matches PICK_MIN_HEIGHT used before the parameter was exposed; sim can lower it because the plane segmenter sometimes picks a wrong plane off the house mesh.",
-            ),
```
JLDominguezM, line 49: The launch argument pick_min_height is gone, so any existing ros2 launch pick_and_place pick_and_place.launch.py pick_min_height:=X invocations in scripts, docs, the wiki, or CI configs will start failing silently as unknown arguments. Worth a quick grep over docs/, scripts/ and any external launch wrappers to make sure nothing references it.

JLDominguezM (changes requested):

manipulation/packages/arm_pkg/config/sensors_3d.yaml
Fernando94654: ok

manipulation/packages/pick_and_place/launch/pick_and_place.launch.py
Fernando94654: There are no invocations of this parameter in scripts, docs, the wiki, or CI configs.

manipulation/packages/pick_and_place/pick_and_place/pick_server.py
Fernando94654: These false positives are no longer a problem

manipulation/packages/perception_3d/src/masive_testin.cpp
Fernando94654: The remove_plane_client is still actively used in the place workflow via place_service_callback. The cleanup only removed it from the pick workflow. In the place flow, the plane is not added to MoveIt's planning scene as a collision object, the extracted plane cloud is passed directly to the heatmap service to compute the optimal placement point on the table surface, so the self-filtering concern doesn't apply here.

manipulation/packages/pick_and_place/pick_and_place/pick_server.py
Fernando94654: This verification is not necessary because the octomap already detects collision objects, and both MoveIt and GPD reject collision grasps.

JLDominguezM (approved):
LGTM

PR #984, opened 2026-05-23
docs: update README so current and new HOME members can add themselves as contributors

htovarm7:
## Description
The goal of this PR is to serve as a reference so that current and new @home team members can easily add themselves to the contributors list


GilMM27 (changes requested):
😂️ fix title and description

PR #987, opened 2026-05-23
Manip/error handler

Fernando94654:
Added safeguard node: 
- Restores the xArm when the emergency stop is pressed, the mode or state is changed, or the boundary limits are invalid; also added respawns.

https://github.com/user-attachments/assets/913f1d68-e94a-4ca4-addf-d8239ba2114e

https://github.com/user-attachments/assets/9dfb8361-5122-434b-b011-616405964bc9


JLDominguezM (dismissed):
Good work but ...
Add some videos of this working

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
```diff
@@ -0,0 +1,269 @@
...
+
+        self.create_subscription(
+            RobotMsg,
+            XARM_ROBOT_STATES_TOPIC,
+            self._on_arm_state,
+            10,
+            callback_group=self.callback_group,
+        )
+
+        self._estop_pub = self.create_publisher(Bool, ESTOP_TOPIC, 10)
```
JLDominguezM, line 64: default qos here is VOLATILE, so anything that subscribes after you already published the estop (a respawn, a late launch) misses it and assumes it's safe to move. probably want TRANSIENT_LOCAL on the publisher and the matching subscribers.

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
```diff
@@ -0,0 +1,269 @@
...
+        return True
+
+    def _switch_controller(self, deactivate: list, activate: list, label: str) -> bool:
+        req = SwitchController.Request()
+        req.deactivate_controllers = deactivate
+        req.activate_controllers = activate
+        req.strictness = SwitchController.Request.BEST_EFFORT
+        future = self.switch_controller_client.call_async(req)
+        start = self.get_clock().now()
+        while not future.done():
```
JLDominguezM, line 311: no sleep in this loop — burns a thread at 100% until the timeout fires. `_call_svc` above already does the `sleep_for(0.1)` thing, can we reuse the same pattern here?

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
```diff
@@ -0,0 +1,269 @@
...
+        self._switch_controller(
+            ["xarm6_traj_controller"], [], "deactivate xarm6_traj_controller"
+        )
+        return self._switch_controller(
+            [], ["xarm6_traj_controller"], "activate xarm6_traj_controller"
+        )
+
+    def _motion_enable(self) -> bool:
+        req = SetInt16ById.Request()
+        req.id = 8
```
JLDominguezM: the 8 = all joints in xarm right? same with `state == 4`, `state in (3,4)`, `state in (1,2)`, `mode == 1` further down. worth moving these to named constants in manipulation_constants.py — that file already holds the topic/service names so it fits.

manipulation/packages/frida_motion_planning/frida_motion_planning/motion_planning_server.py
```diff
@@ -183,13 +186,19 @@ def __init__(self):
...
+            Trigger,
+            MANIPULATION_ENSURE_ARM_READY_SERVICE,
             callback_group=self.callback_group,
         )
 
-        self.current_mode = -1
+        self.create_subscription(
+            Bool,
+            ESTOP_TOPIC,
+            lambda msg: setattr(self, "_in_estop", msg.data),
```
JLDominguezM, line 195: this one *follows* the topic so `_in_estop` flips back to False as soon as the safeguard clears. but pick/place/pour latch True until execute_callback resets it. which one is the intended behavior? mixing the two will confuse someone (probably me in two weeks).

manipulation/packages/frida_motion_planning/frida_motion_planning/motion_planning_server.py
```diff
@@ -649,34 +678,20 @@ def set_gripper_state_callback(self, request, response):
...
-            self.get_logger().info("xarm6_traj_controller reactivated successfully")
-        else:
-            self.get_logger().error("Failed to reactivate xarm6_traj_controller")
+            if (self.get_clock().now() - start).nanoseconds > 15e9:
+                self.get_logger().error("ensure_arm_ready timed out")
+                return
+            self.get_clock().sleep_for(rclpy.duration.Duration(seconds=0.1))
 
+    def reset_xarm_controller(self, request, response):
+        self._call_ensure_arm_ready()
```
JLDominguezM: returning `response` without setting `success`/`message` leaves the caller with no way to tell if the recovery worked. can we propagate the result from `_call_ensure_arm_ready` (or change it to return a bool)?

manipulation/packages/arm_pkg/launch/frida_driver.launch.py
```diff
@@ -92,6 +92,8 @@ def launch_setup(context, *args, **kwargs):
         executable="xarm_driver_node",
         output="screen",
         emulate_tty=True,
+        respawn=True,
```
JLDominguezM, line 95: is respawn=True safe on the xarm driver? if it dies mid-trajectory and the safeguard immediately tries to re-arm, the arm can end up in a weird state. at minimum worth a comment spelling out the assumption — or have the safeguard detect a fresh respawn and require operator confirmation before re-arming.

JLDominguezM (changes requested):
left a few inline. the qos one and the busy-wait are the ones I'd block on, rest are nits.

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
Fernando94654: Default QoS is used across multiple publishers in the home2 repository and works as expected.

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
Fernando94654: nice catch, I added

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
Fernando94654: ok, good practice

manipulation/packages/frida_motion_planning/frida_motion_planning/motion_planning_server.py
Fernando94654: You're right. We standardized on follow-topic (msg.data) across all nodes  pick, place, and pour now match motion_planning_server and manipulation_core. Also removed the manual self._estop = False resets in execute_callback since they're no longer needed. The safeguard is the single source of truth.

manipulation/packages/frida_motion_planning/frida_motion_planning/motion_planning_server.py
Fernando94654: Good point. Changed _call_ensure_arm_ready to return bool

manipulation/packages/arm_pkg/launch/frida_driver.launch.py
Fernando94654: ok, added comment

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
```diff
@@ -0,0 +1,409 @@
...
+            and not self._joints_out_of_bounds()
+        )
+        response.success = ok
+        response.message = "" if ok else "arm not ready after recovery attempt"
+        return response
+
+    def _on_joint_goal_target(self, msg: JointState):
+        """Store the incoming joint target so _ensure_arm_ready can use it when OOB."""
+        target = {name: pos for name, pos in zip(msg.name, msg.position)}
+        self._pending_target_angles = [
```
JLDominguezM, line 132: this cache never expires, if an action publishes a goal, finishes (or fails for some unrelated reason), and the arm later goes out of bounds (OOB) during something completely different, the safeguard will drive to that stale target. could the publisher send an empty JointState (or a separate clear topic) when the action ends, or could we expire this after N seconds?

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
```diff
@@ -0,0 +1,409 @@
...
+        Leaves the arm in mode 0 — caller is responsible for reinit and recovery."""
+        self.get_logger().warn("Normalizing joints via direct xarm control (mode 0)...")
+
+        deltas = [0.0] * len(self._arm_state.angle)
+        for i in bad_indices:
+            name = _JOINT_NAMES[i]
+            lower, upper = JOINT_POSITION_LIMITS[name]
+            raw = self._arm_state.angle[i]
+            delta = 0.0
+            while raw + delta > upper + _BOUNDS_TOLERANCE:
```
JLDominguezM: these two while loops only converge cleanly when the joint range is ≥ 2π. some xarm6 joints (j5 in particular) have a range < 2π  if raw lands in a region where no multiple of 2π falls inside the limits, the first loop drives delta below lower, the second drives it back above upper, and they oscillate forever. at minimum a safety counter + an error log so we don't spin here, and probably bail out of _normalize_joints when no valid delta exists.

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
```diff
@@ -0,0 +1,409 @@
...
+            req = SetInt16.Request()
+            req.data = 0
+            self._call_svc(self._set_state_client, req, 5.0, "set_state(0)")
+
+        if needs_normalize:
+            if self._pending_target_angles is not None:
+                self._normalize_joints_to_target(self._pending_target_angles)
+                self._pending_target_angles = None
+            else:
+                self._normalize_joints(bad_joints)
```
JLDominguezM, line 389: the constant comment for XARM_POSITION_MODE says it bypasses MoveIt, which means these normalization moves don't go through collision checking. if the arm is OOB while holding an object or close to the table, we can drive it straight into something. worth at least a loud warning before the move and a check for attached collision objects or refuse to auto-normalize when there are any.

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
```diff
@@ -0,0 +1,409 @@
...
+                delta += 2 * math.pi
+            deltas[i] = delta
+            self.get_logger().warn(
+                f"{name} OOB ({raw:.4f} rad) — moving by {delta:.4f} rad "
+                f"to {raw + delta:.4f} rad (limits [{lower:.4f}, {upper:.4f}])"
+            )
+
+        req_mode0 = SetInt16.Request()
+        req_mode0.data = XARM_POSITION_MODE
+        self._call_svc(
```
JLDominguezM: _call_svc returns bool, but the result is dropped on the floor here. if set_mode(0) or the following set_state(0) fail, the  set_servo_angle call below is sent in the wrong mode and either silently does nothing or moves unexpectedly. should bail out (and log) when either returns False. same thing applies in _normalize_joints_to_target.

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
JLDominguezM: lgtm

JLDominguezM (changes requested):

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
Fernando94654: That is no neccessary because motion planning server always send the current goal before any request

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
Fernando94654: okey, errors are handled now

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
Fernando94654: Correct. For joints with range < 2π, the 2π step in loop1 can overshoot the entire valid zone; loop2 then pushes it back above upper with no detection. Fix:  analytical solution n = argmin|n| s.t. lower ≤ raw + n·2π ≤ upper, returns None if no valid n exists.

manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
Fernando94654: yes, this is the goal

JLDominguezM (approved):

PR #994, opened 2026-05-23
HRIC improvements

GilMM27:
* Call `deactivate_face_recognition` since we verified it doesn't affect face recognition and improves performance
* Uncommented take bag step
* Don't ask for bag placed confirmation
* Query moondream multiple times for person description. Once for each attribute until 4 are retrieved
* Call `answer_with_context` to generate a sentence with the attributes


task_manager/task_manager/subtask_managers/vision_tasks.py
```diff
@@ -946,11 +947,57 @@ def count_objects(self, object: str) -> tuple[int, list[str]]:
...
+        def handle_result(status, answer):
+            if status == Status.EXECUTION_SUCCESS and answer:
+                cleaned = answer.strip()
+                if cleaned and cleaned[0] != "0":
+                    state["parts"].append(cleaned)
+                    state["count"] += 1
+                seen.add(state["index"])
+
+            while state["index"] in seen and len(seen) < len(attributes):
+                state["index"] = np.random.randint(0, len(attributes))
```
JLDominguezM: If the moondream query returns a non-success status, seen.add(state["index"])is skipped, so the index is never marked as seen. Thewhileloop below won't advance the index and we re-query the same failing attribute forever, never calling the final callback. Moveseen.add(...) outside the success branch (or add a per-attribute attempt cap) so failures still advance.

task_manager/scripts/hric_task_manager.py
```diff
@@ -140,7 +140,28 @@ def timeout(self, timeout: int = 2):
         time.sleep(timeout)
 
     def set_description(self, status, description: str):
```
JLDominguezM, line 142: This description is filled in through a chain of two async callbacks (describe_person --> answer_with_context), but the FSM advances immediately. By the time guest_1 description is read (TAKE_BAG / INTRODUCTION), the chain may not have completed, producing an empty/None description in the spoken sentence. Is there anything guaranteeing the description is ready before it's consumed?

task_manager/scripts/hric_task_manager.py
```diff
@@ -140,7 +140,28 @@ def timeout(self, timeout: int = 2):
...
+                _, natural = f.result()
+                natural_answer = natural if natural else description
+                self.guests[FIRST_GUEST_IDX].description = natural_answer
+                Logger.info(
+                    self, f"Description for {self.guests[FIRST_GUEST_IDX].name}: {natural_answer}"
+                )
+
+            future.add_done_callback(callback)
+        else:
+            self.guests[FIRST_GUEST_IDX].description = description
```
JLDominguezM, line 164: set_description now hardcodes FIRST_GUEST_IDX instead of get_current_guest(). This only works because describe_person is currently called for the first guest only. Consider passing the guest/index explicitly or asserting the assumption, so this doesn't silently break if a second guest ever gets described.

task_manager/task_manager/subtask_managers/vision_tasks.py
```diff
@@ -946,11 +947,57 @@ def count_objects(self, object: str) -> tuple[int, list[str]]:
...
         Logger.info(self.node, "Describing person")
-        prompt = "Briefly describe 4 attributes of the person in the image and only say the description: They are .... (Make sure to mention 4 attributes). For example: shirt color, clothes details, hair color, hair style, if the person has glasses, etc. Don't mention if the person is looking directly at the camera."
-        self.moondream_query_async(prompt, query_person=True, callback=callback)
+        attributes = [
+            "describe the type of clothes and color the person in the front is wearing (dress, pants and shirt, skirt and shirt, shorts and shirt, etc.). Format: The person is wearing [description].",
+            "does the person in the front have glasses? (if not, respond only '0', if yes, describe the glasses).",
+            "does the person in the front have facial hair (if not, respond only '0', if yes, describe the facial hair).",
+            "describe the age group the person in the front belongs to (teenager, young adult, middle-aged, or elderly).",
+            "is the person in the front wearing jewerly? (if not, respond only '0', if yes, describe the jewelry).",
```
JLDominguezM: jewelry**

task_manager/task_manager/subtask_managers/vision_tasks.py
```diff
@@ -946,11 +947,57 @@ def count_objects(self, object: str) -> tuple[int, list[str]]:
...
+            "describe the hair style and color of the person in the front (straight, wavy, curly, in a pony tail, etc.). Format: The person has [description] hair.",
+            "what is the build of the person in the front? (thin, average, athletic, robust, etc.). Format: Their build is [description].",
+        ]
+
+        seen = set()
+
+        state = {"index": np.random.randint(0, len(attributes)), "count": 0, "parts": []}
```
JLDominguezM: Pulling in numpy just for randint is heavy, random.randint(0, len(attributes) - 1) from the stdlib would do the same. Also note np.random.randint is half-open[0, len) while random.randint is inclusive adjust the bound if you switch.

task_manager/task_manager/subtask_managers/hri_tasks.py
```diff
@@ -1241,21 +1241,34 @@ def get_hand_items(self, query: str) -> tuple[list[HandItem], list[HandItem]]:
         by_description = [HandItem(**item) for item in json.loads(raw[1])]
         return by_name, by_description
 
-    # TODO: Make async
     @service_check("llm_wrapper_service", (Status.SERVICE_CHECK, ""), TIMEOUT)
-    def answer_with_context(self, question: str, context: str) -> str:
+    def answer_with_context(
```
JLDominguezM, line 1260: The async callback calls f.result()directly; if the service call fails this raises inside the executor and the 
outer Future never gets a result, hanging any waiter. Consider setting an error status on exception (same applies to the existing extract_data pattern).

JLDominguezM (dismissed):

task_manager/scripts/hric_task_manager.py
GilMM27: no, I was planning on adding it later, but it will MOST likely be ready by the time it is needed

task_manager/scripts/hric_task_manager.py
GilMM27: we only need this for the first guest

task_manager/task_manager/subtask_managers/vision_tasks.py
GilMM27: Fixed the infinite loop: In the handle_result function, I moved seen.add(state["index"]) outside of the if status == Status.EXECUTION_SUCCESS and answer: block. This ensures that the
      attribute index is marked as evaluated even if the query fails or returns an empty answer, effectively preventing the script from infinitely retrying the same attribute.

task_manager/task_manager/subtask_managers/vision_tasks.py
GilMM27: I don't believe it's that expensive but alright

task_manager/task_manager/subtask_managers/hri_tasks.py
GilMM27: added try block

GerardoFJ (approved):
LGTM

PR #996, opened 2026-05-23
Fix tune hotwords clean

CamilaTite26:
I closed the previous PR because the branch got broken and ended up with a lot of merge conflicts. In this new PR I uploaded the cleaned-up code, improved the voice recognition by injecting the initial_prompt into Whisper, and added the new test_hric_hotwords function so everything can be tested quickly without running the full state machine.


frida_constants/frida_constants/hri_constants.py
```diff
@@ -92,3 +94,26 @@ class KNOWLEDGE_TYPE(Enum):
...
+    from ament_index_python.packages import get_package_share_directory
+
+    objects_path = os.path.join(
+        get_package_share_directory("frida_constants"), "data", "objects.json"
+    )
+    with open(objects_path, "r") as f:
+        data = json.load(f)
+        HRIC_DRINK_HOTWORDS = data.get("categories", {}).get("drink", [])
+except Exception:
+    HRIC_DRINK_HOTWORDS = ["water", "coke", "coffee", "tea", "milk", "juice"]
```
GilMM27: this json is used for PPC, but HRIC rules state that the drink may be any popular american drink. I'd rather we just wrote the names here instead of reading them from the json

GilMM27 (changes requested):
I don't think we should add options to the hric call because that forces the answer to be one of those. Instead of modifying hri_tasks.py, just change hric_task_manager.py and add the values to the call you deem appropriate

task_manager/task_manager/subtask_managers/hri_tasks.py
GilMM27, line 1: I rather we didn't auto populate the initial_prompt or context.
Also why did you remove the and not options check? I believe the previous logic made sense

GilMM27 (changes requested):

PR #997, opened 2026-05-23
HRI update download ei model

biweep863:
- write api key to .env
- always download model

### DIFF
--- a/docker/hri/run.sh
+++ b/docker/hri/run.sh
@@ -45,6 +45,7 @@ fi
 setup_common_env "hri" "compose/.env"
 
 [ "$DOWNLOAD_MODEL" == "true" ] && bash ./scripts/download-model.sh
+[ "$TASK" = "--download-model" ] && exit 0
 
 # Create dirs with current user to avoid permission problems
 mkdir -p ../../hri/packages/speech/assets/downloads/offline_voice/model/ \
--- a/docker/hri/scripts/download-model.sh
+++ b/docker/hri/scripts/download-model.sh
@@ -190,33 +190,35 @@ if ask_for_model ei-kws 7; then
     download_ei_model "kws" "${EI_API_KEY_KWS:-}" "1338"
 fi
 
-# Detect available image
-if docker images | grep -q "dustynv/ollama"; then
-    IMAGE="dustynv/ollama:0.6.8-r36.4"
-    COMMAND="ollama serve"
-elif docker images | grep -q "ollama/ollama"; then
-    IMAGE="ollama/ollama"
-    COMMAND=""
-else
-    echo "Error: No compatible Ollama image found. Pulling the default image..."
-    docker pull ollama/ollama:latest
-    IMAGE="ollama/ollama"
-    COMMAND=""
+if ask_for_model qwen3 3 || ask_for_model nomic-embed-text 4; then
+    # Detect available image
+    if docker images | grep -q "dustynv/ollama"; then
+        IMAGE="dustynv/ollama:0.6.8-r36.4"
+        COMMAND="ollama serve"
+    elif docker images | grep -q "ollama/ollama"; then
+        IMAGE="ollama/ollama"
+        COMMAND=""
+    else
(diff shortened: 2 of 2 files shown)


docker/hri/run.sh
```diff
@@ -45,6 +45,7 @@ fi
 setup_common_env "hri" "compose/.env"
 
 [ "$DOWNLOAD_MODEL" == "true" ] && bash ./scripts/download-model.sh
+[ "$TASK" = "--download-model" ] && exit 0
```
GilMM27, line 48: why do you want to exit if the user wanted to download a model?

docker/hri/scripts/download-model.sh
```diff
@@ -91,13 +97,31 @@ download_ei_model() {
...
+                kws)  env_var="EI_API_KEY_KWS"  ;;
+                *)    env_var="EI_API_KEY_$(echo "$model_name" | tr '[:lower:]' '[:upper:]')" ;;
+            esac
+            if grep -q "^${env_var}=" "$ENV_FILE"; then
+                sed -i "s|^${env_var}=.*|${env_var}=${api_key}|" "$ENV_FILE"
+            else
+                printf '\n%s=%s\n' "$env_var" "$api_key" >> "$ENV_FILE"
+            fi
+            echo "API key saved to $ENV_FILE"
+        fi
```
GilMM27, line 113: this is to save the api key after the first time? Explain the user flow

GilMM27 (changes requested):

docker/hri/run.sh
biweep863: after downloading the model, all the contianers start and does not allow downloading ei models again, so you had to run a --stop or --down flag

docker/hri/scripts/download-model.sh
biweep863: yes, this save the api key in the .env, and you could rerun the downloadmodel script and won't ask for the api key again.

docker/hri/run.sh
GilMM27: wdym, that works for me

docker/hri/scripts/download-model.sh
GilMM27: so then when the user needs to change api key they would have to modify the .env? I guess that's fine but defeats the purpose of having this code, I would just rather the user needing to add the .env manually instead

docker/hri/run.sh
biweep863: mmm, i think it is just skipping the download, that is why it does not fail, because it usees the ei container to download the model, so when force to download always the model, it says that the container is being used.

docker/hri/scripts/download-model.sh
biweep863: maybe we could add another menu for selecting modifying it or don't, and would allow the user to not move the .env, but the code could look a little dirty, what do you think would be more suitable?

docker/hri/run.sh
GilMM27: well fine, big trust

docker/hri/scripts/download-model.sh
GilMM27: I'd rather just leave it as simple as possible and add documentation on how to setup the ei

biweep863: so submenu or not?

GilMM27: so just leave it as it was. Always ask for the api key

GilMM27: the only thing I wanted was for the download to not be skipped if the model was already downloaded

biweep863: i fixed that, and improve the flow 🤑

GilMM27: so will it always ask for the api key or are we leaving it like this?

biweep863: the API key does not change with new models, but if necessary the user should change it in the .env.

GilMM27 (approved):
we ball

PR #999, opened 2026-05-24
Auto-compile VAMP Python bindings on manipulation container startup

JLDominguezM:
## Summary

- The `vamp` Python bindings (`_core_ext.*.so`) are a non-standard build artifact — the normal `colcon build` does **not** produce them. After a clean build, a fresh clone, or a platform change (x86 ↔ aarch64) the `.so` goes missing and `vamp_server.py` / the `vamp` planner pipeline fail with `ModuleNotFoundError: No module named 'vamp._core._core_ext'`.
- `docker/manipulation/run.sh` now runs `setup_vamp.sh` as part of the container SETUP chain (both with and without `--build`). It short-circuits in ~1s when the `.so` already exists and compiles it (~1–2 min) only when missing.
- The call is wrapped (`... || echo WARN`) so a build failure is **non-fatal**: the shell/launch still proceeds because the planner's OMPL fallback works without VAMP, rather than locking the user out of the container.

## Why submodule checkout is *not* automated here

The `vamp` submodule source is checked out the same way as every other submodule in the repo — a one-time `git submodule update --init --recursive` after clone or after a pull that adds it. Special-casing only `vamp` would be inconsistent with `gpd`, `pymoveit2`, `xarm_ros2`, `mujoco_ros2_control`. The genuinely VAMP-specific need is the **`.so` compile** (a build step no other submodule has), which is what this PR automates. The manual submodule step and the two distinct `ModuleNotFoundError` symptoms are documented in the plugin ReadMe.

## Test plan

Validated on x86 laptop with the running `home2-manipulation` container, exercising the exact command the SETUP chain generates:

- [x] **`.so` present** → `setup_vamp.sh` short-circuits in **0.34s** ("VAMP already compiled with frida_real") — negligible startup overhead.
- [x] **`.so` missing** → detects ("No compiled .so found"), compiles in **46.5s**, copies it, verifies `import vamp` → "VAMP ready — frida_real with 68 spheres".
- [x] **Failure path** → with a non-existent setup script, the `|| echo` fires and the `&&` chain still reaches the RUN step (no lock-out).
- [x] `import vamp` works after rebuild: `OK spheres: 68`.

## Files

| File | Change |
|---|---|
| `docker/manipulation/run.sh` | Add `VAMP_SETUP` to the SETUP chain (both build/non-build branches), wrapped for non-fatal failure |
| `manipulation/packages/vamp_moveit_plugin/ReadMe.md` | New "Setup / first run" section: submodule one-time step + auto-compile behavior + the two ModuleNotFoundError symptoms |

### DIFF
--- a/docker/manipulation/run.sh
+++ b/docker/manipulation/run.sh
@@ -24,11 +24,14 @@ GPD_EXPORT="export GPD_INSTALL_DIR=/workspace/install/gpd"
 SOURCE="if [ -f install/setup.bash ]; then source install/setup.bash; fi"
 COLCON="colcon build --symlink-install --packages-up-to manipulation_general xarm6_ikfast_plugin xarm_utils vamp_moveit_plugin --packages-ignore realsense_gazebo_plugin xarm_gazebo frida_interfaces"
 CYCLONE_SOURCE="source /usr/local/bin/cyclonedds_setup.sh"
+# Build VAMP's _core_ext.*.so if missing (~1s no-op when already built); the normal colcon build above doesn't produce it.
+# Non-fatal: keep going on failure since the plugin's OMPL fallback works without VAMP.
+VAMP_SETUP="(bash /workspace/src/docker/manipulation/setup_vamp.sh || echo '[WARN] VAMP setup failed — vamp planning unavailable, OMPL fallback still works')"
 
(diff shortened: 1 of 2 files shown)


Fernando94654 (approved):
lgtm!

PR #1010, opened 2026-05-26
Face trt

Fernando94654:
Run InsightFace on TensorRT GPU, fix Dockerfile (it was not detecting the correct onnxruntime-gpu version), and use a Docker volume to avoid reinstalling the model every time.
before:
image
after:
image


htovarm7 (approved):
Excellent finding the bottleneck!

DanaeSG (approved):
lgtm

PR #1013, opened 2026-05-27
ppc display

CamilaTite26:
i also implemented the necessary changes in the run.sh because i could see that when launching the hric flag the display wasn't recreated with the implemented changes, now check if there are changes to build the image

### DIFF
--- a/docker/display/run.sh
+++ b/docker/display/run.sh
@@ -31,9 +31,8 @@ if [ "$ENV_TYPE" != "cpu" ]; then
   add_or_update_variable .env "RUNTIME" "nvidia"
 fi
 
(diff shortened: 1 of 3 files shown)


CamilaTite26:
image

docker/display/run.sh
GilMM27, line 1: I'd rather we left the building to the user explicitly with the flag

hri/packages/display/display/app/ppc/page.tsx
GilMM27, line 1: big trust 🫡️

GilMM27 (changes requested):

biweep863 (approved):
good

GilMM27 (approved):

PR #1016, opened 2026-05-27
Hri merge ei

GilMM27:
image

Merged both EI nodes into one.  It's launched twice with different config
No more infinite error logs after timeout


hri/packages/speech/config/door_eim.yaml
```diff
@@ -0,0 +1,14 @@
+door_eim:
+  ros__parameters:
+    audio_topic: /hri/rawAudioChunk
+    KEYWORD_TOPIC: /hri/speech/ei_detection
```
biweep863, line 4: why don't use REPLACE here??

hri/packages/speech/config/kws_eim.yaml
```diff
@@ -1,14 +1,14 @@
 kws_eim:
   ros__parameters:
-    EI_SERVER_URL: "http://localhost:1338"
-    PROCESSED_AUDIO_TOPIC: REPLACE
+    ei_server_url: "http://localhost:1338"
+    audio_topic: /hri/processedAudioChunk
```
biweep863, line 4: both yaml should have the same variables?

hri/packages/speech/scripts/ei_audio_node.py
```diff
@@ -19,45 +20,34 @@
...
         self.declare_parameter("sample_rate", 16000)
         self.declare_parameter("window_size_s", 1.0)
-        self.declare_parameter("hop_ratio", 0.25)
+        self.declare_parameter("hop_ratio", 1.0)
+        self.declare_parameter("sensitivity_threshold", 0.7)
+        self.declare_parameter("detection_cooldown", 0.0)
         self.declare_parameter("audio_gain", 1.0)
         self.declare_parameter("noise_label", "noise")
-        self.declare_parameter("SAVE_DETECTION_AUDIO", False)
+        self.declare_parameter("save_detection_audio", False)
```
biweep863, line 37: wh ythhe change in names? should be same as hri_constants

hri/packages/speech/scripts/ei_audio_node.py
```diff
@@ -66,42 +56,47 @@ def __init__(self):
...
-            f"({self.hop_size_samples / self.sample_rate * 1000:.0f}ms) | "
-            f"threshold={self.sensitivity_threshold}"
-        )
-
+        self.window_samples = int(self.sample_rate * window_size_s)
+        self.hop_samples = max(1, int(self.window_samples * hop_ratio))
         self.audio_buffer = np.array([], dtype=np.int16)
 
-        self.publisher = self.create_publisher(String, keyword_topic, 10)
+        self.publisher = self.create_publisher(String, result_topic, 10)
```
biweep863, line 87: if the topic will change for keywords as result_topic, maybe in hri_constants could change too

hri/packages/speech/scripts/ei_audio_node.py
```diff
@@ -119,26 +115,27 @@ def _wait_for_server(self, poll_interval: float = 5.0, max_retries: int = 60):
...
                 ).astype(np.int16)
-            self._classify(window)
 
-    def _classify(self, window: np.ndarray):
-        """Send audio window to EI inference server and process results."""
-        features = window.astype(float).tolist()
+            self.classify(window)
+
+    def classify(self, audio_window: np.ndarray):
+        features = audio_window.astype(float).tolist()
```
biweep863, line 138: so this does not cause any overlap errors??

biweep863 (changes requested):
great improvement

hri/packages/speech/scripts/ei_audio_node.py
GilMM27: No, the server should only respond with one inference

hri/packages/speech/scripts/ei_audio_node.py
GilMM27: I'll leave it like this for consistency

hri/packages/speech/config/door_eim.yaml
GilMM27: because kws and ei_detection should be published to different topics since each is handled separately

hri/packages/speech/scripts/ei_audio_node.py
GilMM27: it's just the name of the var, to avoid the discrepancy between only keywords or only events (door)

biweep863 (approved):

PR #1024, opened 2026-05-31
Hri benchmark nlp

biweep863:
Benchmark different models with --benchmark flag in hri area
Change current model to qwen3-4b, best model from benchmark

image


docker/hri/scripts/benchmark-models.txt
GilMM27, line 1: why isn't this a json?

docker/hri/scripts/download-model.sh
GilMM27, line 1: why the reduction in parameters?

run.sh
GilMM27, line 1: remove diff

hri/packages/nlp/benchmark/tasks.py
GilMM27, line 1: this should use the tasks that we already created. If not, changes in one place will have to be replicated here

hri/packages/nlp/benchmark/Dockerfile
GilMM27, line 1: I'm not a fan of having benchmark related files inside the NLP ROS package considering that it isn't related programmatically

hri/packages/nlp/benchmark/config.yaml
GilMM27, line 1: comments in english

docker/hri/run.sh
```diff
@@ -9,17 +9,24 @@ ENV_TYPE="${*: -1}"
...
+while [[ $i -lt ${#ARGS[@]} ]]; do
+  arg="${ARGS[$i]}"
   case "$arg" in
     "--download-model") DOWNLOAD_MODEL="true" ;;
     "--regenerate-db")  REGENERATE_DB="true" ;;
     "--build-proto")    BUILD_PROTO="true" ;;
+    "--benchmark")        BENCHMARK="true" ;;
+    "--benchmark-delete") BENCHMARK="true"; BENCHMARK_PASSTHROUGH+=("--delete") ;;
   esac
+  (( i++ )) || true
```
GilMM27: I'd rather we avoided bash magic and kept it as simple as possible.

docker/hri/scripts/run-benchmark.sh
GilMM27, line 1: 💀️

GilMM27 (changes requested):
we just want to compare some models, this benchmark looks really complete but almost out of place.

docker/hri/scripts/benchmark-models.txt
biweep863: true, let me change it

docker/hri/scripts/download-model.sh
biweep863: it is faster than using 8B and has nearly the same accuracy.

run.sh
biweep863: ok, mb

hri/packages/nlp/benchmark/tasks.py
biweep863: but this will require integration for running benchmarks? a model could last more than 10 minute to download, i was thinkg more of this approach to test new models in bulk, so the logs could be saved and in the orinserver you could leave the benchmark running, and then return after an hour with results. so it was more "automatically" than you have to download models, then start running the test cases in integration and comparing. But if this tests are the "correct" way of doing it, let me change it, but i think running benchmarks will be more tedious.

hri/packages/nlp/benchmark/Dockerfile
biweep863: so where shoudl i put them? in docker/hri or where should we put the benchmarks?

hri/packages/nlp/benchmark/config.yaml
biweep863: mb

docker/hri/run.sh
biweep863: mmm i could change the --benchmark-delete, but i think it is realy easy to run ./run.sh hri --benchmark, but it could be just the bash script as it is in main

docker/hri/scripts/run-benchmark.sh
biweep863: ?

biweep863: 🤪

hri/packages/nlp/benchmark/tasks.py
GilMM27: you could always just run integration on the orinserver too and yes, I think recycling the code is the better approach while maintaining this ease of running

GilMM27 (approved):
I don't understand the cases changes in test hri manager or why they're in this pr but if that looks good to you ok. 👨‍🦯

PR #1026, opened 2026-06-01
fix Dockers build image

GilMM27:
no more infinite transferring context and exporting to image wait times.

Copy only specific packages in integration's Dockerfile to take more advantage of layer caching

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,48 @@
+# Docker build context exclusions
+
+# Source Control
+.git
+.gitmodules
+.githooks
+
+# Build & Cache
+**/build/
+**/install/
+**/log/
+**/devel/
+**/.vscode/
+**/__pycache__/
+**/.pytest_cache/
+**/.ruff_cache/
+**/.clj-kondo/
+**/.lsp/
+**/baml_client/
+
+# Node.js
+**/node_modules/
+
+# ML / Heavy Caches
+**/trt_cache/
+**/moondream_cache/
+**/chromadb/
+**/.cache/
+*.pt
+*.pth
+*.engine
+*.onnx
+*.whl
+
+# Large Assets (Mounted via volumes, not needed in image)
+hri/packages/nlp/assets/**
+!hri/packages/nlp/assets/Modelfile
+!hri/packages/nlp/assets/NewModelfile
+
+# SQL dumps
+docker/hri/sql_dumps/
+
+# Temporary / OS files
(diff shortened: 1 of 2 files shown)


.dockerignore
```diff
@@ -0,0 +1,48 @@
...
+*.whl
+
+# Large Assets (Mounted via volumes, not needed in image)
+hri/packages/nlp/assets/**
+!hri/packages/nlp/assets/Modelfile
+!hri/packages/nlp/assets/NewModelfile
+
+# SQL dumps
+docker/hri/sql_dumps/
+
```
biweep863, line 42: why this? nlp and embedding are not in integration? why this paths?

.dockerignore
```diff
@@ -0,0 +1,48 @@
...
+# ML / Heavy Caches
+**/trt_cache/
+**/moondream_cache/
+**/chromadb/
+**/.cache/
+*.pt
+*.pth
+*.engine
+*.onnx
+*.whl
```
biweep863, line 33: and all this? why this folder, i think when removing this line "COPY ./ /workspace/src", you don't need this right?

docker/integration/Dockerfile
```diff
@@ -15,13 +15,13 @@ RUN baml-cli generate
 
 WORKDIR /workspace
 
-COPY ./ /workspace/src
-RUN sudo apt update && sudo rosdep init && rosdep update
-RUN rosdep install --from-paths ./src/task_manager --ignore-src -y -r
+COPY task_manager /workspace/src/task_manager
+COPY frida_constants /workspace/src/frida_constants
+COPY frida_interfaces /workspace/src/frida_interfaces
```
biweep863, line 20: looks good the change, but i think all dockerfiles, has the full repo in src/ folder, why integration will be the exception? with the .dockerignore, it will easily copy all the project in src isn't it?

biweep863 (commented):
looks good

.dockerignore
GilMM27: I'm ignoring everything heavy that doesn't need to be available for building docker images. It's useful for all areas

docker/integration/Dockerfile
GilMM27: it will still have it since it is a volume, this step is needed for the rosdep install, afterwards it is deleted

biweep863 (approved):
ok lgtm

PR #1032, opened 2026-06-04
Fix hri mock

GilMM27:
didn't really check the mocked responses, just want to not return an error for now


Oscar-gg (approved):
End of an era 🪦🫡

PR #1036, opened 2026-06-10
Make is coherent check more permissive

GilMM27:
It was returning false on a bunch of commands from the command generator


Oscar-gg (approved):
Lgtm

PR #1043, opened 2026-06-16
Navigation OmniBase integration

GerardoFJ:
Omnibase with all nav requirements integration.

-Custom selection of old base with new one
-Nav 2D available

-Tests
-Navigation with task manager integrations


JLDominguezM (approved):
LGTM 👍

PR #1045, opened 2026-06-17
Fix zed offset

Champtiago:
- ZED origins derived from real mount/CAD geometry (no hand-tuning).
- URDF visual matches the real camera position.
- No hardcoded/static offsets left compensating for a wrong URDF.
- Confirmed no ZED offset on the new omni base.
- Functional perception data remains accurate and unaffected.

Captura de pantalla_2026-05-26_11-08-12


Champtiago:
Video on issue, couldnt upload here

robot_description/frida_description/urdf/zed/zed.xacro
```diff
@@ -2,6 +2,7 @@
 <robot xmlns:xacro="http://www.ros.org/wiki/xacro">
 
 <xacro:property name="mesh_path" value="file://$(find frida_description)/urdf/meshes/Gripper/Commercial/" />
+<xacro:property name="black_rgba" value="0.102 0.102 0.102 1" />
```
Fernando94654: why this property?

Champtiago: property from original xacro, not needed in new xacro structure

PR #1047, opened 2026-06-17
Generalize flat object picking and move the estimator to a service

JLDominguezM:
Two changes to how flat objects (cutlery, plates) are picked.

Replaces the three hardcoded cutlery lists (estimator `target_classes`, `PickManager.is_flat_object`, `pick_server.is_flat`) with a single `FLAT_OBJECT_NAMES` constant in `frida_constants`. Placement category stays separate from grasp strategy, and cutlery is a subset so existing picks do not change.

Moves the `flat_grasp_estimator` from the enable/disable topic flow to an `EstimateFlatGrasp` service that collects a few frames, medians the position and returns one pose, so `PickManager` no longer toggles it or averages on its own.

`frida_interfaces` needs a rebuild for the new srv. Tested that it builds (x86 and l4t) and the service comes up. End to end pick on the arm is still pending.


frida_constants/frida_constants/manipulation_constants.py
Fernando94654, line 1: Did you test pick rim?

Fernando94654, line 1: Can you add a demo video?

manipulation/packages/pick_and_place/pick_and_place/tests/test_flat_grasp_routing.py
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
```
Fernando94654: This test is not useful for the home2 repo, it isn't wired into any test runner (no pytest/CI), it asserts a copy of the routing logic rather than the real is_flat_grasp, and it only restates static constants; please delete it.

task_manager/task_manager/subtask_managers/manipulation_tasks.py
```diff
@@ -367,38 +363,10 @@ def pick_object(
...
-        Enables the flat grasp estimator, performs the pick, then disables it."""
-        self.set_flat_grasp_estimator(True)
-        try:
-            result = self.pick_object(object_name)
-        finally:
-            self.set_flat_grasp_estimator(False)
-        return result
+        """Thin alias over pick_object. PickManager auto-detects flat objects
+        (FLAT_OBJECT_NAMES) and gets a pose from the estimate_flat_grasp service."""
+        return self.pick_object(object_name)
```
Fernando94654: please delete pick_cutlery and use pick_object in tasks managers

Fernando94654 (changes requested):

Fernando94654 (approved):
lgtm

PR #1052, opened 2026-06-20
fix urdf

Fernando94654:
Fix real-robot TF tree: drop the static world→base_footprint URDF joint (kept only for MuJoCo) and switch the EKF to publish odom→base_footprint, making   base_footprint the consistent root and fixing the MoveIt frame mismatch that broke picks.


GerardoFJ (approved):
LGTM 🚀

PR #1055, opened 2026-06-20
Robust shelf pick: per-level detection and compartment collision avoidance

JLDominguezM:
Shelf picks detected only from table_stare, which frames the lower levels, so objects on higher shelves weren't found, and the grasp approach/retract collided with the compartment ceiling.

- Detect at each shelf level and pick from the level where the object is found.
- On shelves, prefer a frontal grasp; a top-down approach hits the ceiling. Falls back to the unfiltered set if none qualify.
- After grasping, retract straight out toward the robot (dynamic distance) before moving away, so the arm clears the compartment. Table picks unchanged.

Tested on the robot at all three levels (yogurt, cereal, coke) without collision.


GerardoFJ (approved):
LGTM 🚀

PR #1058, opened 2026-06-21
Add omnidriver update with on-MCU EKF support

Roger412:
The omnidriver dashboard is now current with home-custom-base's odrive_comm reference implementation: consumes the on-MCU 6-state EKF (quaternion + covariance), broadcasts odom -> base_link TF (replacing robot_localization for the omnibase config), parses the firmware's queued error-reporting stream into a dedicated log panel, and supports arbitrary ODrive parameter writes over CAN (RxSdo) for tuning without reflashing. ESTOP that stops the motors and sets them to IDLE while pressed was also added.

Full design detail (EKF math, firmware error codes, CAN protocol, per-task flow diagrams) lives in home-custom-base: https://github.com/RoBorregos/home-custom-base/tree/main — see firmware/STM32H7_OMNIBASE_CAN_BNO085/omnibase_documentation.md.

Notes:
  - robot_localization should not be launched if tf is enabled in the driver    
  - The dashboard's Axis Lab tab (per-motor manual control/calibration) is
    untested against real hardware.
  - enable_web_gui defaults to False for this package; the dashboard's ROS
    topics/TF still publish normally with it off.
  - ESTOP untested with FRIDA.


GerardoFJ (approved):
LGTM 🚀

PR #1060, opened 2026-06-21
nav distance query test gpsr

Isai-JQ:
Adds a way to ask how far point A → point B is along a real nav2-planned path, without moving the robot. GPSR/HRI can use this distance to make decisions before committing to a navigation goal.


GerardoFJ (approved):
LGTM :rocket:

PR #1062, opened 2026-06-21
Add shelf height calibrator

JLDominguezM:
Adds a shelf height calibrator.

`ros2 run xarm_utils shelf_height_calibrator.py`: click each cabinet shelf in RViz
(Publish Point), average the clicks per level, transform to base_link Z, and write
`config/shelf_levels.json` (versioned, loaded via package share).

Tested on the robot: produced the three base_link heights now committed in `shelf_levels.json`.

The FSM loader that reads these heights was split out of this PR (it touches the task
manager) and will land with the related task-manager work.


GerardoFJ (approved):
LPTM :rocket: "Looks papu to me"

PR #1064, opened 2026-06-22
Vision fixes

Fernando94654:
- Fix navigation areas in GPSR and HRIC commands
- Set face recognition to be deactivated by default
- Use cached YOLO TensorRT models in the MoonDream node


GerardoFJ (approved):
LGTM :rocket:

PR #1066, opened 2026-06-22
Make manipulation safeguard opt-in (default off)

JLDominguezM:
## What
Gate the `manipulation_safeguard` node behind a launch arg `enable_safeguard` (default `false`).

## Why
The safeguard over-triggers and halts the arm mid-trajectory (servo motor 5 error + a latched e-stop on `/manipulation/estop`), which silently blocks normal picks. Making it opt-in keeps picks running by default while the node stays available.

## Enable
`ros2 launch pick_and_place pick_and_place.launch.py enable_safeguard:=true`

No behavior change when enabled.


GerardoFJ (approved):
LGTM 🚀 🚀 🚀 🚀 🚀 🚀 🚀  GOAT

PR #1067, opened 2026-06-22
Add per-level shelf pick and vision confirmation

JLDominguezM:
Per-level shelf detection for cabinet picks (scan a level, detect, then pick) and vision confirmation by re-detection for pick and place. Uses the RViz-calibrated shelf heights.

Validated on the robot via the keyboard pick/place flow.


Isai-JQ (approved):
LGTM 🚀

PR #1068, opened 2026-06-22
Improve pick robustness: faster moves and cartesian shelf retract

JLDominguezM:
Manipulation robustness, validated on the robot:
- Faster picks: drop the per-move ensure_arm_ready 5s stall when the safeguard is off
- Cartesian shelf retract that pulls straight toward the robot, retimed for control
- Auto-clear stale collision objects before each pick so grasp goal sampling isn't blocked
- Cutlery grasp robustness, frontal grasp filter, and a pick benchmark tool


GerardoFJ (approved):
LPTM :rocket: (Looks papu to me)

PR #1071, opened 2026-06-23
Make manipulation safeguard a global node-level toggle

JLDominguezM:
The `enable_safeguard` flag only existed as a launch argument in `pick_and_place.launch.py`, so the safeguard kept running under `hric` and any other launch that starts the node. This moves the toggle into the node so a single setting governs it everywhere.

`manipulation_safeguard` now reads `enable_safeguard`, defaulting from the `FRIDA_ENABLE_SAFEGUARD` env var (default `false`). When disabled it does not broadcast the autonomous e-stop and does not run the recovery timer, but it still serves `ensure_arm_ready`, so `motion_planning_server` keeps working without waiting on a missing service or losing on-demand recovery.

The launch-level `enable_safeguard` condition is removed so the node always starts and self-gates; keeping it would double-gate and drop `ensure_arm_ready`.

Enable with `FRIDA_ENABLE_SAFEGUARD=true` or `-p enable_safeguard:=true`.

### DIFF
--- a/manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
+++ b/manipulation/packages/manipulation_general/manipulation_general/manipulation_safeguard.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 
 import math
+import os
 import rclpy
 from rclpy.node import Node
 from rclpy.action import ActionClient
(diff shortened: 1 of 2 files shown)


GerardoFJ (approved):
LGTM

PR #1074, opened 2026-06-24
Robust shelf pick and place

JLDominguezM:
Makes the shelf pick and place work reliably across the three cabinet levels.

Pick: fix the ZED mesh origin in the URDF (it was about 3cm forward and caused
false camera-shelf collisions that blocked deep frontal grasps), add a shelf GPD
config so grasps come in from the front, flip the grasp so the camera faces up,
and hold the level pose through the pick.

Place: pull the place toward the front edge so the arm and camera don't jam the
compartment ceiling on a deep reach, and add an explicit ceiling collision before
the return, since the eye-in-hand camera never looks up and the octomap has no
record of it.

Tested on hardware on levels 1, 2 and 3.


GerardoFJ (approved):
LGTM

PR #1075, opened 2026-06-24
Place laundry

Fernando94654:
Add place pose for place clothes in doing laundry(without obstructing the camera)
(The gripper doesn't work in the video xd)

https://github.com/user-attachments/assets/201cc8fa-5b08-4918-94ea-367dfd7163bd


GerardoFJ (approved):
LPTM :rocket: (Looks papu to me)

PR #1080, opened 2026-06-26
Feat/arm aim nav goal

Isai-JQ:
Adds a nav_goal_arm_pointer node that yaws the xArm6 (joint 1 only) so the arm continuously points at the current Nav2 destination while the base drives, then returns the arm to face the base front once navigation ends.


GerardoFJ (approved):
LGTM 🚀

PR #1085, opened 2026-06-26
fix of the nodes

Isai-JQ:
When navigation ends, the arm pointer no longer chases joint1 by degrees with velocity control (imprecise). It now releases    velocity mode, hands the arm back to MoveIt, and sends it to the named NAV_POSE via the MoveJoints action (/manipulation/move_joints_action_server), guaranteeing it always ends facing front. It publishes /nav/arm_ready=True only once the
action completes, so nav_central keeps waiting for the arm to arrive before returning success.

xArm mode-contention guard: the pointer attempts to take velocity mode once per goal; if it fails (manipulation owns the arm), it warns and does not retry — eliminating the storm of Failed to set xArm mode 4 and the thrashin


GerardoFJ (approved):
LGTM :rocket:

PR #1086, opened 2026-06-26
Arm pointer: yield to manipulation on the xArm + keep gripper  closed on mode switches

Isai-JQ:
Motion_planning_server (the sole xArm owner) publishes   /manipulation/arm_busy (latched Bool), True while a 
 MoveJoints/MoveToPose goal executes, cleared in finally.  - nav_goal_arm_pointer subscribes and yields while busy: it never 
 touches /xarm/set_mode, stops commanding, and relinquishes so nav doesn't wait on it. It only takes velocity mode when the arm is free. The handoff is clean because the existing manipulation_safeguard.ensure_arm_ready already reinits MoveIt mode before each manipulation move, so the pointer never has to  set MoveIt mode back itself.
Gripper:
On startup the pointer calls /xarm/config_tgpio_reset_when_stop  with 0, disabling the tool-GPIO reset on state changes so the gripper stays put across mode switches.

### DIFF
--- a/frida_constants/frida_constants/manipulation_constants.py
+++ b/frida_constants/frida_constants/manipulation_constants.py
@@ -38,6 +38,10 @@ JOINTN_VELOCITY_MODE = 4
 
 MOVE_JOINTS_ACTION_SERVER = "/manipulation/move_joints_action_server"
 ESTOP_TOPIC = "/manipulation/estop"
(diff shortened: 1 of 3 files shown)


GerardoFJ (approved):
LGTM :rocket:

PR #1122, opened 2026-07-02
Eliminate ambiguity for color clothes

DanaeSG:
### Summary

Fixed an issue in HRI attribute matching where invalid clothing values could incorrectly resolve to `unknown` from the Pose and Gesture enum. For example, when receiving an attribute such as `color "x"` and `x` was not recognized as a valid clothing item, the nearest match could be `unknown`, resulting in incorrect behavior.

### Changes

* Removed `unknown` as a candidate value during attribute matching.
* Added `color` as a valid attribute category so color-related queries are correctly identified and processed as color matching requests.
* Improved color extraction logic to consider only the first color token. This prevents ambiguous matches in cases such as `"black blues"` (intended as `"black blouse"`), ensuring:

  * `black` is interpreted as the color.
  * `blues/blouse` is matched independently as the clothing item.

### Pending HRI Update

* Verify whether the current task already requires approaching the target person and avoid unnecessary prompts such as "approach me" when the robot should navigate toward the person.


GilMM27 (approved):
looks good, but we should test it first

PR #1217, opened 2026-08-18
Dynamic build constants

### LINKED ISSUE #1162: Frida constants
Improve how we manage constants. Currently we build constants with frida_interfaces, this takes a lot of time for just changing one constant.

Isai-JQ:
* add build_frida_constants.sh (symlink-install + dereference build-only symlinks)
* wire script into cpu/cuda/l4t compose files
* leave frida_interfaces build untouched
* verify idempotent across rebuilds, all 3 environments
* requires ./run.sh --clean-interfaces once to migrate existing cache


biweep863:
point it towards int-migration, and please add captures of changing a constant and crashing, then another picture, where you return it to its natural state and works. You could do the test with hri or any other area, because it will trhow the error that a constant was not found.

Isai-JQ:
Retargeted to `int-migration` (merged it in, resolved the compose conflicts keeping the Jazzy paths + the script). Diff is still the same 4 files.

Test on the Orin (`frida-jazzy`, l4t): fresh clone of this branch, `./run.sh frida_interfaces` once, then hri container with the same mounts as `hri-ros.yaml`. No `colcon build` and no `./run.sh frida_interfaces` between any of the runs below.

**1. Change the value of a constant → hri uses the new value.** `SPEAK_SERVICE` edited in `src/`, `say.py` restarted, `ros2 service list` shows the new name; restore with `git checkout`, old name is back:
3-constant-value-changed-no-rebuild


**2. Remove/rename a constant → hri crashes with "constant not found"** (proves it reads `src/` live, not a stale build; note the install file is a symlink to `src/`):
1-constant-changed-crash


**3. Restore it → hri works again:**
2-constant-restored-works

biweep863 (approved):
lgtm, change target branch to main 🚀

Isai-JQ:
PR #1217 now targets main instead of int-migration. GitHub shows it as mergeable with clean status: no conflicts.

PR #1226, opened 2026-08-25
Base class for vision_general's per-task nodes

### LINKED ISSUE #428: Package re-structure
Some scripts that are not so general should be added to their own packages. For example: face recognition, tracker, etc. Essentially we would want to keep only the task_comands scripts in the vision_general package

Fernando94654:
Added Base class for vision_general's per-task nodes

### DIFF
--- a/vision/packages/vision_general/scripts/task_nodes/vision_runtime.py
+++ b/vision/packages/vision_general/scripts/task_nodes/vision_runtime.py
@@ -1,0 +1,196 @@
+"""Shared base for vision_general's per-task nodes.
+
+Counterpart to object_detector_2d's `base_detector_node.py`: the camera
+subscriptions, the ``/vision/<name>/active`` gate, camera-rotation tracking, a
+TF buffer, an optional ``DebugImagePublisher`` and a ``call_service`` that waits
+on an event instead of spinning.
+
+Subclasses opt in through the constructor flags and read ``self.image``,
+``self.depth_image`` and ``self.camera_info`` exactly as they did when each node
+owned its own subscriptions.
+"""
+
+import threading
+
+import rclpy
+import rclpy.qos
+import tf2_ros
+from cv_bridge import CvBridge, CvBridgeError
+from frida_constants.vision_constants import (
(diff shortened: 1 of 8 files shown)


vision/packages/vision_general/scripts/task_nodes/vision_runtime.py
```diff
@@ -0,0 +1,196 @@
...
+        if debug_topic:
+            self.debug_publisher = DebugImagePublisher(
+                self,
+                debug_topic,
+                debug_name or node_name,
+                max_hz=debug_max_hz,
+                callback_group=self.callback_group,
+            )
+
+        self.active = True
```
m2galindo: why not to add an explicit default? Node start  True until other nodes set them to False, which can cause accidental activations

vision/packages/vision_general/scripts/task_nodes/vision_runtime.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def call_service(
+        self, client, request, timeout: float = 5.0, name: str = "service"
+    ):
+        """Call a service, waiting on an event. Returns the response, or None on timeout."""
+        if not client.service_is_ready():
+            self.get_logger().warn(f"{name} unavailable")
+            return None
+
+        future = client.call_async(request)
```
m2galindo: Should we check for future.result() exceptions to log errors, and add a fallback if the executor blocks?

m2galindo (commented):
Leave some comments

vision/packages/vision_general/scripts/task_nodes/vision_runtime.py
Fernando94654: Actually, this method is not used, I removed

vision/packages/vision_general/scripts/task_nodes/vision_runtime.py
Fernando94654: I added a method attribute with default = True

m2galindo (approved):
LGTM

m2galindo (approved):
LGTM

PR #1248, opened 2026-09-01
1196 rescue the unmerged laundry nav work

### LINKED ISSUE #1196: Rescue the unmerged laundry nav work
- Check the code of move_relative in laundry_robocup_final and decide if we adopt it, change it or drop it. Right now there is no way to move the robot a short distance without going through nav2 planning, so it probably helps us. If we take it, it goes in its own PR because it is clean.

- Check the code of wall_aligner that is implemented in doing papu2 and decide if it stays as its own node or if it becomes part of the general approach we planned. 

- Check align_parallel because it overlaps with table_docker. Say which one we keep or if they solve different problems.

- Check set_obstacle_avoidance because it turns off the lidar and zed obstacle layers on both costmaps. It has no auto restore, so if a task turns it off and crashes the robot drives blind. Do not merge it as it is. Also find out why laundry needed it, probably the basket shows up as an obstacle in front of the robot, and that would be a lead for the laundry sticking issue.

JaredAlPa:
Recover basket-approach logic from doing_laundry (align left/right in approach_point)


CamilaTite26 (changes requested):
Good job! Please try to test it after merging the PR for moving DashGo, and then rebase this branch so it includes the files from that PR.
This approach is for the laundry task, so please test it by running the entire task I could help you with that.

CamilaTite26 (changes requested):
Also, if this implementation helps with the doing laundry task, I think we should also recover the implementation code from that task manager. Please also check whether this implementation was applied to any other task managers.

frida_interfaces/navigation/srv/ApproachPoint.srv
```diff
@@ -5,6 +5,8 @@
 # there through the standard goal pipeline.
 geometry_msgs/PointStamped target
 float32 standoff        # meters from the target; <= 0 uses the default (0.65)
+string align            # "", "face", "left" or "right" (see above)
+float32 final_dsistance  # align left/right only: final base_link<->target gap (m); <= 0 skips the close-in strafe
```
CamilaTite26, line 9: I think there is a typo. The service is called final_distance.

JaredAlPa: Fixed, thanks for catching that.

JaredAlPa:
> Also, if this implementation helps with the doing laundry task, I think we should also recover the implementation code from that task manager. Please also check whether this implementation was applied to any other task managers.

Done, I recovered the code from the laundry task manager and checked the other task managers.

PR #1264, opened 2026-09-07
Change Zed frame constant to new name

Fernando94654:
- Change Zed frame constant to new name
- Remove harcoded use


m2galindo (approved):
LGTM

PR #1284, opened 2026-09-17
Stt test programatically

### LINKED ISSUE #1011: test STT programatically
Mainly for .wav to see if it's necessary to edit the pipeline

TEC-Andres:
# Summary 
Programmatic STT testing on benchmarks. 

Measures **accuracy** (Word Error Rate) and **latency** (Real-Time Factor) of faster-whisper models on pre-recorded test audio.

> [!NOTE]
> Although using the same approaches from #1034. It was remade from scratch.

## Interactive runner

```bash
chmod +x run.sh
./run.sh                          # interactive menu
./run.sh --model distil-large-v3  # single model
./run.sh --all                    # all models in models.json
./run.sh --tasks accuracy         # accuracy only
./run.sh --tasks latency          # latency only
./run.sh --runs 5                 # custom run count
```

## Recording new test audio

Requires PulseAudio. Records 5 seconds at 16 kHz mono:

```bash
pip install sounddevice scipy
python record.py recordings/new_command.wav
```

## Adding test cases

Edit `test_cases.json`. Each entry:

```json
{
  "name": "test_name",
  "audio_file": "filename.wav",
  "expected_transcript": "expected text",
  "gain": 2.0
}
```

## Results

CSV and JSON reports are saved to `results/`.

## DEMO
|   |   |
| --- | --- |
| image | image |
| **Basic Usage:** Running tiny.en locally | **Hotwords:** Running tiny.en using hotwords (lower RFT) |

## Dependencies

```
faster-whisper
ctranslate2>=4.0.0
numpy
sounddevice     # for record.py only
scipy           # for record.py only
```


hri/benchmarks/stt/run.sh
```diff
@@ -0,0 +1,121 @@
...
+sys.path.insert(0, '.')
+import report as rpt
+results = json.loads('''$ALL_RESULTS''')
+rpt.print_model_table('$model_name', results)
+"
+
+    # Save JSON
+    TS=$(date +%Y%m%d_%H%M%S)
+    REPORT_PATH="$RESULTS_DIR/${model_name}_${TS}.json"
+    echo "$ALL_RESULTS" | jq --arg ts "$(date -Iseconds)" '{"timestamp": $ts, "models": {"'$model_name"'}} + .' > "$REPORT_PATH"
```
luisrosales852: I think there's something wrong here - the quoting on `{"models": {"'$model_name"'}}` is unbalanced, the script doesn't even parse:

```
$ bash -n run.sh
run.sh: line 121: unexpected EOF while looking for matching `"'
run.sh: line 122: syntax error: unexpected end of file
```

and even once the quotes balance, `{"models": {"distil-large-v3"}}` isn't valid json (key with no value). have you tried just doing it with `--arg`? something like

```bash
jq --arg ts "$(date -Iseconds)" --arg m "$model_name" '{timestamp: $ts, models: {($m): .}}'
```

can we get a `bash -n` (or shellcheck) on this in CI so it doesn't happen again

hri/benchmarks/stt/run.sh
```diff
@@ -0,0 +1,121 @@
...
+
+        RESULT=$(cd "$SCRIPT_DIR" && python3 -c "
+import json, sys
+sys.path.insert(0, '.')
+from tasks import TASK_REGISTRY
+task_cls = TASK_REGISTRY.get('$task')
+if task_cls is None:
+    print(json.dumps({'error': 'unknown task $task'}))
+    sys.exit(0)
+r = task_cls.run(model_name='$model_name', runs=$RUNS)
```
luisrosales852: this passes `model_name=` but `tasks.py` declares `run(model: str, runs: int = 1, **kwargs)`, so it lands in kwargs and `model` is never filled:

```
TypeError: AccuracyTask.run() missing 1 required positional argument: 'model'
```

and since the traceback gets captured by `2>&1` into `$RESULT`, the `jq --argjson` on the next line chokes on it and `set -e` kills the run. so even with the quoting fixed above this dies on the first task. just `model='$model_name'` should do it

hri/benchmarks/stt/run.sh
```diff
@@ -0,0 +1,121 @@
...
+
+        ALL_RESULTS=$(echo "$ALL_RESULTS" | jq --argjson r "$RESULT" --arg t "$task" '. + {($t): $r}')
+    done
+
+    # Print table
+    cd "$SCRIPT_DIR" && python3 -c "
+import sys, json
+sys.path.insert(0, '.')
+import report as rpt
+results = json.loads('''$ALL_RESULTS''')
```
luisrosales852: interpolating the json blob straight into a python triple-quoted string feels pretty fragile - one apostrophe in a transcript ("what's the time") and this blows up, and we're dumping model output in there. could we pipe it on stdin and `json.load(sys.stdin)` instead? same for `'$model_name'` a couple lines down

hri/benchmarks/stt/run.sh
```diff
@@ -0,0 +1,121 @@
...
+#   ./run.sh --model distil-large-v3  # run accuracy + latency for one model
+#   ./run.sh --all                    # benchmark every model in models.json
+#   ./run.sh --tasks accuracy         # run only accuracy
+#   ./run.sh --tasks latency          # run only latency
+#   ./run.sh --runs 5                 # override run count for latency
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(realpath "$SCRIPT_DIR/../../..")"
```
luisrosales852: nit: `REPO_ROOT` isn't used anywhere. also `realpath` isn't there by default on mac, so if anyone runs this locally it'd fail on a var we don't even need - just drop it

hri/benchmarks/stt/benchmark_stt.py
```diff
@@ -0,0 +1,423 @@
...
+    )
+    parser.add_argument("--language", default="en", help="Language code (default: en)")
+    parser.add_argument(
+        "--runs", type=int, default=3, help="Number of latency runs (default: 3)"
+    )
+    parser.add_argument("--audio", help="Audio file for latency benchmark")
+    parser.add_argument(
+        "--test-cases", default=TEST_CASES_FILE, help="Path to test_cases.json"
+    )
+    parser.add_argument("--no-vad", action="store_true", help="Disable VAD filter")
```
luisrosales852: `--no-vad`, `--language`, `--hotwords` and `--initial-prompt` are all parsed but never read again - `args.language` etc. don't appear anywhere below, and all three call sites (`run_accuracy`, `run_latency`, `run_batch`) go in with just the model. so `--no-vad` silently keeps vad on, which is the kind of thing that'd make us misread a benchmark.

`transcribe_file` already takes all of them, so it's mostly plumbing them through the three runners. hotwords especially I'd want working - "frida" is the word we care most about and whisper loves turning it into "frieda"/"freeda"

hri/benchmarks/stt/benchmark_stt.py
```diff
@@ -0,0 +1,423 @@
...
+                    "passed": False,
+                }
+            )
+            continue
+
+        result = transcribe_file(audio_path, model_name=model_name)
+        actual = result["text"]
+        wer = calculate_wer(expected, actual)
+        accuracy = 1.0 - wer
+        passed = accuracy >= 0.8
```
luisrosales852: not sure 0.8 means much on utterances this short. `stop` is one word so it's pass/fail on exact match, and `hello frida` is two words so a single miss = 0.5 -> fail. effectively the threshold is "exact match" for most of test_cases.json and 0.8 only kicks in on the 4-5 word ones.

might be more useful to just report WER and track it over time rather than pass/fail? or keep pass/fail but make it exact-match-after-normalization so at least it's honest about what it's measuring. wdyt

related: `calculate_wer` can go above 1.0 when the hypothesis is longer than the reference (whisper hallucinating on a 1-word clip will do it), so `accuracy = 1.0 - wer` goes negative and then the avg does too. probably want to clamp

hri/benchmarks/stt/benchmark_stt.py
```diff
@@ -0,0 +1,423 @@
...
+) -> dict:
+    latencies = []
+    rtf_values = []
+    for _ in range(n_runs):
+        result = transcribe_file(audio_path, model_name=model_name)
+        latencies.append(result["processing_time"])
+        if result["real_time_factor"] is not None:
+            rtf_values.append(result["real_time_factor"])
+
+    duration = result["audio_duration"]
```
luisrosales852, line 252: this reads `result` after the loop, so it's relying on the loop variable leaking out. `--runs 0` gives a NameError instead of a clean error. small thing but maybe grab the duration inside the loop, or guard `n_runs < 1` up top.

also should we throw away the first run as a warmup? the model load itself is outside the timer which is good, but the first transcribe is usually noticeably slower (cuda kernels etc) and with `--runs 3` that's a third of the average

hri/benchmarks/stt/benchmark_stt.py
```diff
@@ -0,0 +1,423 @@
...
+        result = transcribe_file(path, model_name=model_name)
+        print(f"  {wav}: {result['text']}  (RTF={result['real_time_factor']})")
+        results.append({"file": wav, **result})
+    return results
+
+
+# ── Reporting ────────────────────────────────────────────────────────────────
+
+
+def print_accuracy_report(results: list[dict], model_name: str) -> None:
```
luisrosales852: we've now got two reporting paths - these functions plus `report.py`, and `report.py` is only ever reached from the inline python in run.sh. they also disagree on what "accuracy" means: here/tasks.py it's `1 - avg WER`, in `report.py` `save_json` it's `passed / len(cases)`. same key, same reports dir, two different numbers.

can we pick one? I'd lean on everything going through `report.py` and this file just being the library

luisrosales852 (commented):
nice, this is a lot cleaner than #1034, thanks for redoing it. left a bunch of comments.

main thing before I can really test it: `run.sh` doesn't run at all for me. `bash -n hri/benchmarks/stt/run.sh` fails with a syntax error on the jq line at the bottom, and even with that fixed the python it shells out to calls the task with the wrong kwarg. so I think only the `benchmark_stt.py` path has actually been exercised so far? the rest is mostly small stuff / questions.

hri/benchmarks/stt/test_cases.json
```diff
@@ -0,0 +1,68 @@
+[
+  {
+    "name": "basic_greeting",
+    "audio_file": "hello_frida.wav",
+    "expected_transcript": "hello frida",
+    "gain": 2.0
```
luisrosales852, line 6: `gain` is in every entry but nothing reads it - `run_accuracy` only touches `audio_file` and `expected_transcript`. either apply it before transcribing or drop it, otherwise someone's going to tweak it and wonder why nothing changes

hri/benchmarks/stt/record.py
```diff
@@ -0,0 +1,32 @@
...
+if len(sys.argv) > 1:
+    target_path = Path(sys.argv[1])
+    if target_path.suffix.lower() != ".wav":
+        target_path = target_path.with_suffix(".wav")
+    OUTPUT_FILE = str(target_path)
+else:
+    OUTPUT_FILE = "hello_frida.wav"
+
+PULSE_SOURCE = "default"
+SAMPLE_RATE = 16000
```
luisrosales852, line 16: heads up, the wavs in `recordings/` don't match what this produces - they're 48kHz and 2-3.5s, this records 16kHz for 5s:

```
bring_water.wav  ch=1 rate=48000 dur=3.48s  333902 bytes
hello_frida.wav  ch=1 rate=48000 dur=2.47s  237646 bytes
stop.wav         ch=1 rate=48000 dur=2.24s  215118 bytes
```

so those weren't made with this script. not broken (whisper resamples to 16k internally anyway) but it's ~3.1MB of binary going into the repo for something that'd be ~1MB at 16k, and the mismatch will confuse whoever adds a test case next. could we re-encode them to 16k mono, and maybe have record.py take the duration as an arg?

hri/benchmarks/stt/report.py
```diff
@@ -0,0 +1,174 @@
...
+                cases = r.get("cases", [])
+                passed = sum(1 for c in cases if c["passed"])
+                total = len(cases)
+                pct = (passed / total * 100) if total else 0
+                row += f"{pct:.0f}%".rjust(16)
+            print(row)
+
+
+def save_json(all_results: dict, output_dir: str) -> str:
+    os.makedirs(output_dir, exist_ok=True)
```
luisrosales852: `results/` gets created at runtime but isn't gitignored, so the first person to run this will have a pile of csv/json in `git status`. `hri/benchmarks/nlp/` handles it with a `results/.gitkeep` - can we match that here (nlp's isn't ignored either tbh, so maybe just add both to .gitignore while you're at it)

hri/benchmarks/stt/tasks.py
```diff
@@ -0,0 +1,85 @@
+"""STT benchmark task definitions.
+
+Each task class wraps a benchmark_stt function and exposes a standard
+``run(model, runs)`` interface so the runner (run.sh / test_hri_manager)
+can invoke any task uniformly.
```
luisrosales852: docstring says test_hri_manager can invoke these but there's no wiring for that in the PR. is that coming in a follow-up? the nlp benchmark goes through `./run.sh integration --test-hri` rather than shelling python out of run.sh, would be nice if stt eventually did the same so we're not maintaining two runner styles

hri/benchmarks/stt/report.py
TEC-Andres: Resolved on [0e6f766](https://github.com/RoBorregos/home2/pull/1284/commits/0e6f766f3a6fdff508cbcf2c79098408c734db59)

TEC-Andres:
Force due to `rebase -i`. Removed .wav files from git history tree. Everything else remains the same.

TEC-Andres:
Removed previous .wav files from git history tree. Required force to keep everything running. Everything else remains the same.

hri/benchmarks/stt/record.py
TEC-Andres: Resolved on [856d073](https://github.com/RoBorregos/home2/pull/1284/commits/856d07308cf93905429b16f095358d3ded8635e6)

hri/benchmarks/stt/test_cases.json
TEC-Andres: Resolved on [46bc95f](https://github.com/RoBorregos/home2/pull/1284/commits/46bc95f77a04a4718b07c974b5029d94e79d976d)

hri/benchmarks/stt/benchmark_stt.py
TEC-Andres: Resolved on [8db644c](https://github.com/RoBorregos/home2/pull/1284/commits/8db644c4e14048e191b4280f8e206e884cb53c99)

hri/benchmarks/stt/benchmark_stt.py
TEC-Andres: Resolved on [a6814e4](https://github.com/RoBorregos/home2/pull/1284/commits/a6814e46b2b03393d4c67a8b35a7bc2da0afac98)

hri/benchmarks/stt/benchmark_stt.py
TEC-Andres: Resolved on [87cf206](https://github.com/RoBorregos/home2/pull/1284/commits/87cf20611af9bc3cdcc3851b97abd0c3babde37b)

hri/benchmarks/stt/run.sh
TEC-Andres: Resolved on [07c1e88](https://github.com/RoBorregos/home2/pull/1284/commits/07c1e888633a503b64bd087f16d0efae989f346b)

hri/benchmarks/stt/run.sh
TEC-Andres: Resolved on [07c1e88](https://github.com/RoBorregos/home2/pull/1284/commits/07c1e888633a503b64bd087f16d0efae989f346b)

hri/benchmarks/stt/run.sh
TEC-Andres: Resolved on [07c1e88](https://github.com/RoBorregos/home2/pull/1284/commits/07c1e888633a503b64bd087f16d0efae989f346b)

hri/benchmarks/stt/run.sh
TEC-Andres: Resolved on [07c1e88](https://github.com/RoBorregos/home2/pull/1284/commits/07c1e888633a503b64bd087f16d0efae989f346b)

hri/benchmarks/stt/tasks.py
TEC-Andres: I'd could leave this as an issue, to wire up hri benchmarks. For now, I'll leave it as is; removing the comment on the process.

hri/benchmarks/stt/benchmark_stt.py
```diff
@@ -0,0 +1,268 @@
...
+        "segments": all_segments,
+        "language": info.language if info else "unknown",
+        "language_probability": round(info.language_probability, 4) if info else 0,
+        "audio_duration": round(duration, 3),
+        "processing_time": round(elapsed, 3),
+        "real_time_factor": round(elapsed / duration, 4) if duration > 0 else None,
+    }
+
+
+def _apply_gain(audio_path: str, gain: float) -> str:
```
biweep863, line 136: i think this functions assumes that it is 16 bit the audio, is not necessarry for this pr, but you oculd make a safe run here, to avoid an unexpected error if 24 or 8 bit audios.

hri/benchmarks/stt/report.py
```diff
@@ -0,0 +1,179 @@
...
+                r = all_results[m].get(task, {})
+                cases = r.get("cases", [])
+                passed = sum(1 for c in cases if c["passed"])
+                total = len(cases)
+                pct = (passed / total * 100) if total else 0
+                row += f"{pct:.0f}%".rjust(16)
+            print(row)
+
+
+def save_json(all_results: dict, output_dir: str) -> str:
```
biweep863: is this function used?

hri/benchmarks/stt/report.py
```diff
@@ -0,0 +1,179 @@
+"""Terminal output and JSON report writer for STT benchmarks."""
```
biweep863: is this file used? i think the run.sh does not run it directly. could you tell me specifically in a short way how is this file used in the benchmark?

biweep863 (changes requested):
looks good, could you add a screenshot or some evidence of this pipeline working in the pr description?

hri/benchmarks/stt/report.py
TEC-Andres: Report is being used on the helper file _print_table.py, which is used directly on run.sh. However, it is kind of a mess

hri/benchmarks/stt/report.py
TEC-Andres: Removed on [c89f89d](https://github.com/RoBorregos/home2/pull/1284/commits/c89f89da3af2cd016ef453f45ec80c7d6071c32c)

hri/benchmarks/stt/report.py
TEC-Andres: Functions and pipeline edited on [4aa534f](https://github.com/RoBorregos/home2/pull/1284/commits/4aa534f044776135d272cc4c5d857eb129336b1e)

hri/benchmarks/stt/benchmark_stt.py
TEC-Andres: Delimited on [dab186a](https://github.com/RoBorregos/home2/pull/1284/commits/dab186aec4392ec4f1ae87708d390f9bff935caf)

biweep863 (approved):
lgtm, just document on the issue your findings

PR #1293, opened 2026-09-18
Clean logs and standarize debug nodes

### LINKED ISSUE #981: clean logs
* add debug flag and standardize it for all nodes (like debug_log fun).
* I only want to see if the one is initializing or if it has finished initializing.
* Exceptions may apply for important information that should be checked before a round.

biweep863:


CamilaTite26 (approved):
lgtm 👍

